PR #35, opened 2018-10-04
Update to RAML 1.0 MODPERMS-47

adamdickmeiss:


dcrossleyau:
I was expecting this to end up like the mod-users, where the RAMLs move out of the shared space.

Is this going to be a two-phase move?

While the rearrangement is happening, probably best to configure the Jenkinsfile to turn off the "publishAPI" and "runLintRamlCop" stages. I can adjust the configuration after the merge to master and when it has settled.

dcrossleyau:
Ah yes, sorry. I was getting myself a bit confused about how to handle the transition of the apidocs configuration.

I did run the lint-raml here locally on this branch, and all is well.

I have switched those jobs off in Jenkinsfile. After the merge and updated config, i will switch them back on.

PR #39, opened 2018-12-17
Modperms 48

kurtnordstrom:


dcrossleyau:
I updated the apidocs configuration api.yml, so the next time the lint-raml job runs it will assess the files in their new location.

PR #65, opened 2019-10-01
Add cache for full permission objects

hjiebsco:
Cache the whole permission table to speed up the expansion of full permission objects. When the cache is stale, return as is to avoid performance hit to current request, but at the same time use a Vertx 1ms delay timer to trigger off cache refresh. Use a flag to avoid multiple concurrent refresh requests.


hjiebsco:
> Looks ok at a glance. When the mod-authtoken hits its own cache, this will not help us much.. IN fact, in some cases make it worse.. However, there are other users of mod-permissions, so it seems sensible to me. I'll look again, when you have increased code coverage in testing (as SQ reports).

Thanks. I will work on Sonar complains. Curious in what cases the perm cache makes it worse and what are the possible reasons?

PR #79, opened 2020-06-02
Missing test cases for MODPERMS-36

adamdickmeiss:
This issue was filed a long time ago and was fixed .. this adds a few tests, so we can close the issue

### DIFF
--- a/src/main/java/org/folio/rest/impl/PermsAPI.java
+++ b/src/main/java/org/folio/rest/impl/PermsAPI.java
@@ -340,7 +340,7 @@ public class PermsAPI implements Perms {
                       getErrorResponse(errStr))));
               return;
             }
-            if (cpfdRes.result()) {
+            if (Boolean.TRUE.equals(cpfdRes.result())) {
               asyncResultHandler.handle(Future.succeededFuture(
                   PutPermsUsersByIdResponse.respond400WithTextPlain(
                       String.format("Cannot add permissions flagged as 'dummy' to users"))));
--- a/src/main/java/org/folio/rest/impl/PermsAPI.java
+++ b/src/main/java/org/folio/rest/impl/PermsAPI.java
@@ -340,7 +340,7 @@ public class PermsAPI implements Perms {
                       getErrorResponse(errStr))));
               return;
             }
-            if (cpfdRes.result() == true) {
+            if (cpfdRes.result()) {
               asyncResultHandler.handle(Future.succeededFuture(
                   PutPermsUsersByIdResponse.respond400WithTextPlain(
                       String.format("Cannot add permissions flagged as 'dummy' to users"))));
@@ -1351,10 +1351,8 @@ public class PermsAPI implements Perms {
         promise.complete(finalMissingPermissions);
       }
     });
-    return promise.future().compose(mapper -> {
-      return findMissingPermissionsFromList(connection, permissionListCopy,
-          vertxContext, tenantId, finalMissingPermissions);
-    });
+    return promise.future().compose(mapper -> findMissingPermissionsFromList(connection, permissionListCopy,
+        vertxContext, tenantId, finalMissingPermissions));
   }
 
   private Future<List<String>> getAllExpandedPermissionsSequential(
@@ -1391,10 +1389,8 @@ public class PermsAPI implements Perms {
           }
         });
 
-    return promise.future().compose(res -> {
-      return getAllExpandedPermissionsSequential(listOfListsCopy, vertxContext,
-          tenantId, res);
-    });
+    return promise.future().compose(res -> getAllExpandedPermissionsSequential(listOfListsCopy,
+        vertxContext, tenantId, res));
   }
 
   private Future<List<String>> getExpandedPermissionsSequential(List<String> permissionList,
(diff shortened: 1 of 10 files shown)


src/test/java/org/folio/permstest/RestVerticleTest.java
```diff
@@ -189,8 +191,42 @@ public void testPermsSeq(TestContext context) {
...
- public void testGroup(TestContext context){
+  @Test
+  public void testPermsUsersPermissionsInvalidUUID(TestContext context)
+      throws InterruptedException, ExecutionException, TimeoutException {
+    String url = "http://localhost:" + port + "/perms/users/123/permissions";
+
+    CompletableFuture<Response> response = new CompletableFuture();
+    send(url, context, HttpMethod.POST, postPermUserPermRequest,
+        SUPPORTED_CONTENT_TYPE_JSON_DEF, 400,  new HTTPResponseHandler(response));
+    Response addPermsResponse = response.get(5, TimeUnit.SECONDS);
```
hjiebsco, line 202: Consider to add this after line 202
`assertEquals(400, addPermsResponse.code);`

src/test/java/org/folio/permstest/RestVerticleTest.java
```diff
@@ -189,8 +191,42 @@ public void testPermsSeq(TestContext context) {
...
+
+  @Test
+  public void testPermsUsersPutInvalidUUID(TestContext context)
+      throws InterruptedException, ExecutionException, TimeoutException {
+    String url = "http://localhost:" + port + "/perms/users/123";
+
+    CompletableFuture<Response> response = new CompletableFuture();
+    send(url, context, HttpMethod.PUT, postPermUsersRequest,
+        SUPPORTED_CONTENT_TYPE_JSON_DEF, 400,  new HTTPResponseHandler(response));
+    Response addPermsResponse = response.get(5, TimeUnit.SECONDS);
```
hjiebsco, line 213: Consider to add this after line 213. Also consider to rename `addPermsResponse` to be PUT related
`assertEquals(404, addPermsResponse.code);`

src/test/java/org/folio/permstest/RestVerticleTest.java
```diff
@@ -189,8 +191,42 @@ public void testPermsSeq(TestContext context) {
...
+
+  @Test
+  public void testPermsUsersDeleteInvalidUUID(TestContext context)
+      throws InterruptedException, ExecutionException, TimeoutException {
+    String url = "http://localhost:" + port + "/perms/users/123";
+
+    CompletableFuture<Response> response = new CompletableFuture();
+    send(url, context, HttpMethod.DELETE, "",
+        SUPPORTED_CONTENT_TYPE_JSON_DEF, 400,  new HTTPResponseHandler(response));
+    Response addPermsResponse = response.get(5, TimeUnit.SECONDS);
```
hjiebsco, line 224: Consider to add this after line 224. Also consider to rename `addPermsResponse` to be DELETE related
`assertEquals(404, addPermsResponse.code);`

src/test/java/org/folio/permstest/RestVerticleTest.java
adamdickmeiss: OMG.. send does not use the errorCode parameter. I'll fix that..

src/main/java/org/folio/rest/impl/PermsAPI.java
```diff
@@ -140,1165 +140,1094 @@ private static void report(String noise, Logger logger) {
...
-      });
+      }
     } catch (Exception e) {
-      String errStr = "Error running vertx on context:" + e.getLocalizedMessage();
+      String errStr = "Error running vertx on context:" + e.getMessage();
       logger.error(errStr, e);
-      asyncResultHandler.handle(io.vertx.core.Future.succeededFuture(
-        GetPermsUsersResponse.respond500WithTextPlain(getErrorResponse(errStr))));
+      asyncResultHandler.handle(Future.succeededFuture(
+          GetPermsUsersResponse.respond500WithTextPlain(errStr)));
```
hjiebsco, line 195: Seems that `getErrorResponse()` was removed after the refactoring.

src/main/java/org/folio/rest/impl/PermsCache.java
```diff
@@ -59,15 +58,15 @@ private PermsCache() {
...
       if (ar.succeeded()) {
-        future.complete(ar.result().expandPerms(perms));
+        promise.complete(ar.result().expandPerms(perms));
       } else {
-        future.fail(ar.cause());
+        promise.fail(ar.cause());
       }
     });
-    return future;
+    return promise.future();
```
julianladisch, line 69: This should be replaced by map, no need to manually handle the fail case:

```
return getPermCache(vertxContext, tenantId, new HashSet<>(perms)).map(permCache -> permCache.expandPerms(perms));
```

src/main/java/org/folio/rest/impl/PermsCache.java
```diff
@@ -79,15 +78,15 @@ private PermsCache() {
...
       if (ar.succeeded()) {
-        future.complete(ar.result().getFullPerm(permissionName));
+        promise.complete(ar.result().getFullPerm(permissionName));
       } else {
-        future.fail(ar.cause());
+        promise.fail(ar.cause());
       }
     });
-    return future;
+    return promise.future();
```
julianladisch, line 89: Should be replaced by map as above.

PR #82, opened 2020-06-16
Honor id/UUID when posting permission set + refactor MODPERMS-84

adamdickmeiss:
Fixes the issue of MODPERMS-84. It also returns 400 error where it previously returned 500 in a number of cases where user input . eg tenant is bad and fails due to bad user input. Not an internal error.. The 500 error handling has been re-done.. and so the coverage is low for this PR. Still overall coverage is up from 63.7 to 71.1.

### DIFF
--- a/src/main/java/org/folio/rest/impl/PermsAPI.java
+++ b/src/main/java/org/folio/rest/impl/PermsAPI.java
@@ -10,12 +10,12 @@ import io.vertx.core.json.JsonArray;
 import io.vertx.core.json.JsonObject;
 import io.vertx.core.logging.Logger;
 import io.vertx.core.logging.LoggerFactory;
+import io.vertx.ext.web.RoutingContext;
 import java.util.ArrayList;
 import java.util.Base64;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
-import java.net.URLDecoder;
 import javax.ws.rs.core.Response;
 import java.util.Objects;
 import java.util.stream.Collectors;
@@ -26,9 +26,9 @@ import org.folio.rest.jaxrs.model.PermissionListObject;
 import org.folio.rest.jaxrs.model.PermissionNameListObject;
 import org.folio.rest.jaxrs.model.PermissionNameObject;
 import org.folio.rest.jaxrs.model.PermissionUser;
-import org.folio.rest.jaxrs.model.PermissionUserListObject;
 import org.folio.rest.jaxrs.model.PermissionUpload;
 import org.folio.rest.jaxrs.resource.Perms;
+import org.folio.rest.persist.PgUtil;
 import org.folio.rest.persist.PostgresClient;
 import org.folio.rest.persist.Criteria.Criteria;
 import org.folio.rest.persist.Criteria.Criterion;
@@ -44,7 +44,6 @@ import org.folio.rest.tools.utils.ValidationHelper;
 import static org.folio.rest.RestVerticle.MODULE_SPECIFIC_ARGS;
 
 /**
- *
  * @author kurt
  */
 @SuppressWarnings("java:S3740")
@@ -73,7 +72,7 @@ public class PermsAPI implements Perms {
     private Operation operation;
 
     public FieldUpdateValues(String fieldValue, String permissionName,
-      PermissionField field, Operation operation) {
+                             PermissionField field, Operation operation) {
       this.fieldValue = fieldValue;
       this.permissionName = permissionName;
       this.field = field;
@@ -99,7 +98,6 @@ public class PermsAPI implements Perms {
 
   private static final String TABLE_NAME_PERMS = "permissions";
   private static final String TABLE_NAME_PERMSUSERS = "permissions_users";
-  private static final String OKAPI_TENANT_HEADER = "x-okapi-tenant";
   private static final String USER_NAME_FIELD = "'username'";
   private static final String USER_ID_FIELD = "'userId'";
   private static final String ID_FIELD = "'id'";
@@ -107,7 +105,7 @@ public class PermsAPI implements Perms {
   private final Logger logger = LoggerFactory.getLogger(PermsAPI.class);
   private static boolean suppressErrorResponse = false;
   private static boolean doExtraReport = Boolean.parseBoolean(MODULE_SPECIFIC_ARGS
-    .getOrDefault("report.extra.logging", "false"));
+      .getOrDefault("report.extra.logging", "false"));
 
   private static CQLWrapper getCQL(String query, String tableName, int limit, int offset) throws FieldException {
     CQL2PgJSON cql2pgJson = new CQL2PgJSON(tableName + ".jsonb");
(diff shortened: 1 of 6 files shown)


src/main/java/org/folio/rest/impl/PermsAPI.java
```diff
@@ -140,51 +139,24 @@ private static void report(String noise, Logger logger) {
 
   @Override
   public void getPermsUsers(int length, int start, String sortBy, String query,
```
julianladisch, line 141: Can you add `@Validate` here and for all other methods?

src/main/java/org/folio/rest/impl/PermsAPI.java
```diff
@@ -613,20 +534,18 @@ public void postPermsUsersPermissionsById(String id, String indexField,
...
       PostgresClient.getInstance(vertxContext.owner(), tenantId).get(
           TABLE_NAME_PERMSUSERS,
           PermissionUser.class, new Criterion(useridCrit), true, false,
           getReply -> {
             try {
               if (getReply.failed()) {
-                logger.error("Error checking for user: " + getReply.cause()
-                    .getMessage(), getReply.cause());
+                String errStr = "Error checking for user: " + getReply.cause().getMessage();
+                logger.error(errStr);
```
julianladisch, line 546:
```suggestion
                logger.error(errStr, getReply.cause());
```

src/main/java/org/folio/rest/impl/PermsAPI.java
```diff
@@ -763,14 +677,15 @@ public void deletePermsUsersPermissionsByIdAndPermissionname(
...
     try {
-      String tenantId = TenantTool.calculateTenantId(okapiHeaders.get(OKAPI_TENANT_HEADER));
+      String tenantId = TenantTool.tenantId(okapiHeaders);
       Criteria idCrit = getIdCriteria(indexField, "=", id);
       PostgresClient.getInstance(vertxContext.owner(), tenantId).get(TABLE_NAME_PERMSUSERS,
           PermissionUser.class, new Criterion(idCrit), true, false, getReply -> {
             if (getReply.failed()) {
-              logger.error("Error checking for user: " + getReply.cause().getMessage());
+              String errStr = getReply.cause().getMessage();
+              logger.error(errStr);
```
julianladisch, line 686:
```suggestion
              logger.error(errStr, getReply.cause());
```

src/main/java/org/folio/rest/impl/PermsAPI.java
```diff
@@ -859,8 +778,9 @@ public void postPermsPermissions(PermissionUpload entity, Map<String, String> ok
       PostgresClient.getInstance(vertxContext.owner(), tenantId).get(
           TABLE_NAME_PERMS, Permission.class, new Criterion(nameCrit), true, false, getReply -> {
             if (getReply.failed()) {
-              logger.error("Error getting existing permissions: " + getReply.cause().getMessage());
-              asyncResultHandler.handle(Future.succeededFuture(PostPermsPermissionsResponse.respond500WithTextPlain("Internal server error")));
+              String errStr = getReply.cause().getMessage();
+              logger.error(errStr);
```
julianladisch, line 782:
```suggestion
              logger.error(errStr, getReply.cause());
```

src/main/java/org/folio/rest/impl/PermsAPI.java
```diff
@@ -892,7 +813,9 @@ public void postPermsPermissions(PermissionUpload entity, Map<String, String> ok
                   if (postReply.failed()) {
                     postgresClient.rollbackTx(connection, done -> {
                       logger.error("Unable to save new permission: " + postReply.cause().getMessage());
```
julianladisch, line 815:
```suggestion
                      logger.error("Unable to save new permission: " + postReply.cause().getMessage(), postReply.cause());
```

src/main/java/org/folio/rest/impl/PermsAPI.java
```diff
@@ -938,33 +859,33 @@ public void getPermsPermissionsById(String id, Map<String, String> okapiHeaders,
...
       idCrit.addField(ID_FIELD);
       idCrit.setOperation("=");
       idCrit.setVal(id);
       PostgresClient.getInstance(vertxContext.owner(), tenantId)
           .get(TABLE_NAME_PERMS, Permission.class, new Criterion(idCrit), true, false, getReply -> {
             if (getReply.failed()) {
-              logger.error("Error in getReply: " + getReply.cause().getMessage());
-              asyncResultHandler.handle(Future.succeededFuture(GetPermsPermissionsByIdResponse.respond500WithTextPlain("Internal server error")));
+              String errStr = getReply.cause().getMessage();
+              logger.error(errStr);
```
julianladisch, line 871:
```suggestion
              logger.error(errStr, getReply.cause());
```

src/main/java/org/folio/rest/impl/PermsAPI.java
```diff
@@ -223,151 +195,120 @@ public void postPermsUsers(PermissionUser entity, Map<String, String> okapiHeade
...
-      String tenantId = TenantTool.calculateTenantId(okapiHeaders.get(OKAPI_TENANT_HEADER));
-      Criteria idCrit = getIdCriteria(indexField, "=", decodedId);
+      String tenantId = TenantTool.tenantId(okapiHeaders);
+      Criteria idCrit = getIdCriteria(indexField, "=", id);
       PostgresClient.getInstance(vertxContext.owner(), tenantId).get(TABLE_NAME_PERMSUSERS, PermissionUser.class,
           new Criterion(idCrit), true, false, queryReply -> {
             if (queryReply.failed()) {
-              String errStr = "queryReply failed: " + queryReply.cause().getMessage();
+              String errStr = queryReply.cause().getMessage();
               logger.error(errStr);
```
julianladisch, line 257:
```suggestion
              logger.error(errStr, queryReply.cause());
```

src/test/java/org/folio/permstest/RestVerticleTest.java
```diff
@@ -696,34 +980,21 @@ public void testGroup(TestContext context) throws InterruptedException, Executio
 
   private void send(String url, TestContext context, HttpMethod method, String content,
                     String contentType, Handler<HttpClientResponse> handler) {
```
julianladisch, line 982: Duplicated code should be extracted into methods for get, post, put, delete. A few examples:
```
private Response send(String tenant, HttpMethod method, String path, String content, TestContext context) {
  try {
    CompletableFuture<Response> futureResponse = new CompletableFuture();
    send(tenant, "http://localhost:" + port + path, context, method, content,
          SUPPORTED_CONTENT_TYPE_JSON_DEF, new HTTPResponseHandler(futureResponse));
    return futureResponse.get(5, TimeUnit.SECONDS);
  } catch (Exception e) {
    context.fail(e);
    return null;
  }
}

private Response get(String path, TestContext context) {
  return send("diku", HttpMethod.GET, path, null, context);
}

private Response get(String tenant, String path, TestContext context) {
  return send(tenant, HttpMethod.GET, path, null, context);
}

private Response put(String path, JsonObject content, TestContext context) {
  return send("diku", HttpMethod.PUT, path, content.encode(), context);
}

private Response put(String path, String content, TestContext context) {
  return send("diku", HttpMethod.PUT, path, content, context);
}

private Response put(String tenant, String path, JsonObject content, TestContext context) {
  return send(tenant, HttpMethod.PUT, path, content.encode(), context);
}

```

adamdickmeiss: Yes. It's quite bad.. Something like restassured would be better. Can we postpone this to MODINVSTOR-83?

julianladisch: Yes, postpone to MODPERMS-83. I don't think that restassured would be better.

PR #84, opened 2020-06-22
Modperms 85 invalid CQL when encoding permission name

adamdickmeiss:
Sorry about the old commits.. used from old master.. There are only two commits since master = 5.11.0 as of today.

There's a calll missing in RMB.. something that takes CQLWrapper and a Connection (PostgresClient.get). Oddly enough there is an update that takes that combo. For this reason there is some redundant code ..  But at least that was better with this commit (you can construct CQLWrapper from Criterion for  later RMBs).


julianladisch:
https://issues.folio.org/browse/RMB-653 : Urldecode path parameters like /perms/users/12%2334

adamdickmeiss:
> https://issues.folio.org/browse/RMB-653 : Urldecode path parameters like /perms/users/12%2334

MODPERMS-85 actual problem is not the one above.. But I saw it and tried to fix it. Note that URL encoding is normally for query parameters.. I'm not sure we should do it for path components.. In fact I think being able to put ANYTHING but non-problematic characters in paths is bad web-service design. In this case, the has indexField as query parameter and value as path-component.. Instead it should just have been a CQL query field=value and we would never have this issue. THings like an UUID can exist just fine as a path component.

PR #85, opened 2020-06-24
MODPERMS-83 Increase code coverage and fix code smells

adamdickmeiss:


src/main/java/org/folio/rest/impl/PermsAPI.java
```diff
@@ -382,15 +380,11 @@ public void deletePermsUsersById(String userid, Map<String, String> okapiHeaders
...
-      Criteria idCrit = new Criteria();
-      idCrit.addField(ID_FIELD);
-      idCrit.setOperation("=");
-      idCrit.setVal(userid);
-      PostgresClient pgClient = PostgresClient.getInstance(vertxContext.owner(),
-          tenantId);
+      Criterion idCrit = getIdCriterion(userid);
+      PostgresClient pgClient = PostgresClient.getInstance(vertxContext.owner(), tenantId);
       pgClient.startTx(connection -> {
         pgClient.get(TABLE_NAME_PERMSUSERS, PermissionUser.class,
```
julianladisch, line 386: should pass connection as first parameter

src/main/java/org/folio/rest/impl/PermsAPI.java
```diff
@@ -428,7 +422,7 @@ public void deletePermsUsersById(String userid, Map<String, String> okapiHeaders
                     return;
                   }
                   pgClient.delete(connection, TABLE_NAME_PERMSUSERS,
-                      new Criterion(idCrit), deleteReply -> {
+                      idCrit, deleteReply -> {
```
julianladisch, line 425:
```suggestion
                      userid, deleteReply -> {
```

src/main/java/org/folio/rest/impl/PermsAPI.java
```diff
@@ -857,12 +851,8 @@ public void getPermsPermissionsById(String id, Map<String, String> okapiHeaders,
                                       Context vertxContext) {
     try {
       String tenantId = TenantTool.tenantId(okapiHeaders);
-      Criteria idCrit = new Criteria();
-      idCrit.addField(ID_FIELD);
-      idCrit.setOperation("=");
-      idCrit.setVal(id);
       PostgresClient.getInstance(vertxContext.owner(), tenantId)
-          .get(TABLE_NAME_PERMS, Permission.class, new Criterion(idCrit), true, false, getReply -> {
+          .get(TABLE_NAME_PERMS, Permission.class, getIdCriterion(id), true, false, getReply -> {
```
julianladisch, line 855: Can PgUtil.getById been used?

src/main/java/org/folio/rest/impl/PermsAPI.java
```diff
@@ -1063,7 +1046,7 @@ public void deletePermsPermissionsById(String id,
                         return;
                       }
                       pgClient.delete(connection, TABLE_NAME_PERMS,
-                          new Criterion(idCrit), deleteReply -> {
+                          idCrit, deleteReply -> {
```
julianladisch, line 1049:
```suggestion
                          id, deleteReply -> {
```

src/main/java/org/folio/rest/impl/PermsAPI.java
```diff
@@ -1089,7 +1072,7 @@ public void deletePermsPermissionsById(String id,
             } else {
               try {
                 PostgresClient.getInstance(vertxContext.owner(), tenantId).delete(TABLE_NAME_PERMS,
-                    new Criterion(idCrit), deleteReply -> {
+                    idCrit, deleteReply -> {
```
julianladisch, line 1075:
```suggestion
                    id, deleteReply -> {
```

src/main/java/org/folio/rest/impl/PermsAPI.java
```diff
@@ -1997,29 +1895,32 @@ private boolean allowAccessByPermission(String permissions, String permissionNam
...
       } else {
         return checkPermlistForDummy(permListCopy, vertxContext, tenantId);
       }
     });
   }
 
-  private Criteria getIdCriteria(String indexField, String operation, String value) {
+  private static Criterion getIdCriterion(String id) {
+    Criteria idCrit = new Criteria();
+    idCrit.addField(ID_FIELD);
```
julianladisch, line 1917: Add `idCrit.setJSONB(false)` to make use of the primary key index resulting in best performance:
https://github.com/folio-org/raml-module-builder/blob/master/domain-models-runtime/src/main/java/org/folio/rest/persist/Criteria/Criteria.java#L30

src/main/java/org/folio/rest/impl/PermsAPI.java
```diff
@@ -1997,29 +1895,32 @@ private boolean allowAccessByPermission(String permissions, String permissionNam
...
+    idCrit.setOperation("=");
+    idCrit.setVal(id);
+    return new Criterion(idCrit);
+  }
+
+  private static Criterion getIdCriterion(String indexField, String value) {
     Criteria crit = new Criteria();
     if (indexField == null || indexField.equals("id")) {
       crit.addField(ID_FIELD);
```
julianladisch, line 1926: Add `idCrit.setJSONB(false)` to make use of the primary key index resulting in best performance:
https://github.com/folio-org/raml-module-builder/blob/master/domain-models-runtime/src/main/java/org/folio/rest/persist/Criteria/Criteria.java#L30

src/main/java/org/folio/rest/impl/TenantPermsAPI.java
```diff
@@ -223,12 +175,12 @@ public void postTenantpermissions(OkapiPermissionSet entity, Map<String, String>
...
+      Iterator<Map.Entry<String, Future<Boolean>>> it = futureMap.entrySet().iterator();
       while (it.hasNext()) {
-        Map.Entry pair = (Map.Entry) it.next();
-        Future<Boolean> existsCheckFuture = (Future<Boolean>) pair.getValue();
-        if (existsCheckFuture.result() == false) {
-          notFoundList.add((String) pair.getKey());
+        Map.Entry<String, Future<Boolean>> pair = it.next();
+        Future<Boolean> existsCheckFuture = pair.getValue();
+        if (Boolean.FALSE.equals(existsCheckFuture.result())) {
+          notFoundList.add(pair.getKey());
```
julianladisch, line 183:
```
      futureMap.forEach( (permName, existsCheckFuture) -> {
        if (Boolean.FALSE.equals(existsCheckFuture.result())) {
          notFoundList.add(permName);
        }
      });
```
or
```
      List<String> notFoundList = futureMap.entrySet().stream()
      .filter(entry -> Boolean.FALSE.equals(entry.getValue().result()))
      .map(Map.Entry::getKey)
      .collect(Collectors.toList());
```

src/main/java/org/folio/rest/impl/PermsAPI.java
adamdickmeiss: Good catch

src/main/java/org/folio/rest/impl/PermsAPI.java
adamdickmeiss: OK. Not that I think it improves readability.. idCrit is the one we are dealing with and when idCrit is using correct query type, it should be the same.

src/main/java/org/folio/rest/impl/PermsAPI.java
adamdickmeiss: Actually, yes. But changes behavior a bit (not for the worse)

src/main/java/org/folio/rest/impl/PermsAPI.java
adamdickmeiss: ok

src/main/java/org/folio/rest/impl/PermsAPI.java
adamdickmeiss: ok

src/main/java/org/folio/rest/impl/PermsAPI.java
adamdickmeiss: Done and ID_FIELD must be unquoted.

src/main/java/org/folio/rest/impl/PermsAPI.java
adamdickmeiss: Yep

src/main/java/org/folio/rest/impl/TenantPermsAPI.java
adamdickmeiss: OK. Took the first one..

src/main/java/org/folio/rest/impl/PermsAPI.java
```diff
@@ -117,10 +114,7 @@ private static CQLWrapper getCQL(String query, String tableName) throws FieldExc
 
   private final Messages messages = Messages.getInstance();
```
julianladisch, line 115: Can you remove this unused variable?

src/main/java/org/folio/rest/impl/PermsAPI.java
```diff
@@ -117,10 +114,7 @@ private static CQLWrapper getCQL(String query, String tableName) throws FieldExc
...
 
-  private String getInternalError(String response) {
-    if (suppressErrorResponse) {
-      return "Internal Server Error: Please contact Admin";
-    }
+  static String getInternalError(String response) {
     return response;
   }
```
julianladisch, line 119: Can you remove this no-op method?

PR #98, opened 2020-09-22
MODPERMS-103: PUT /perms/users/{id} and PUT /perms/permissions/{id} to return 204

julianladisch:
Currently
PUT /perms/users/{id} and
PUT /perms/permissions/{id}
return 200 and the input entity on success.

The client does not need the input entity.
The API doesn't return the metadata property with the new data
created by the database trigger.
Most other FOLIO PUT APIs do not return the entity and use
204 "No content" instead of the 200 HTTP status code.

This pull request changes from 200 to 204 HTTP status code.

Open question:
Is this a breaking change so that we need to bump the interface version
from 5.3 to 6.0 and the mod-permissions version from 5.11.3 to 6.0.0?


hjiebsco (approved):
Maybe bundle this breaking change with the JDK 11 and RMB 31 upgrade in one release?

adamdickmeiss (commented):
I don't this breaking change is worth it. FWIW, Okapi returns 200, which is hardly a normal service. But either 204 or 200 is acceptable in "RESTful" terms.

marcjohnson-kint (commented):
I'm not sure it has been agreed that changing a response status code is a compatibility breaking change.

Given this pull request is treating it as such, are there associated pull requests for every module that depends upon mod-permissions to compensate for that?

julianladisch:
It was agreed that the small change is not worth the breaking change.

PR #106, opened 2020-11-10
MODPERMS-115 Static Permission Migration

craigmcnally:
## Purpose
Cleanup the PoC from [OKAPI-839](https://issues.folio.org/browse/OKAPI-839) and complete implementation.  See  [MODPERMS-115](https://issues.folio.org/browse/MODPERMS-115)

## Approach
* Update _tenantPermissions endpoint to have three lists of permissions - `newPermissions`, `modifiedPermisions`, `removedPermissions`
  * user permissions are updated/removed as needed
  * permissions are added/updated/deleted as needed
  * specify `javaType` in okapiPermission.json so it can be referenced by the three lists mentioned above while keeping a reasonable class name.
  * addition/subtraction of modified permission's subpermissions is not handled here but instead stubbed out.  
* Update unit tests

See https://wiki.folio.org/display/FOLIJET/Migration+of+Static+Permissions+Upon+Upgrade


craigmcnally:
Closing this.  Given the amount that's changed since the POC I thought starting fresh would be easier.  I'll keep this branch around for reference though.  See branch MODPERMS-115_start_fresh

PR #107, opened 2020-12-18
MODPERMS-115 Static Permission Migration

craigmcnally:
## Overview
Implement static permission migration.  

See https://issues.folio.org/browse/MODPERMS-115

## Approach
See https://wiki.folio.org/display/DD/Migration+of+Static+Permissions+Upon+Upgrade

## Scope
* While new code uses promises, this PR does not attempt to update the entire codebase to use promises.  That work will be handled in a separate Story/PR.
* The following cases are not handled in this PR.  These will be addressed in separate stories/PRs.
  * Inactive permissions can be granted to users.  Eventually we want to disallow this.
  * If a module is enabled which references an inactive permission, mod-permissions allows that.  Eventually we want to disallow this and return an error.
 
## Notes
* An earlier PoC was done in branch MODPERMS-115.  Given the changes in the approach since then, I decided to start fresh, but keep that branch around for reference.  This is intended to be the branch which will eventually be merged, not the other one.


ramls/definedByObject.json
```diff
@@ -0,0 +1,31 @@
+{
+  "$schema" : "http://json-schema.org/draft-04/schema#",
+  "title" : "Okapi Permission Object",
```
julianladisch, line 3: Is this name more self-explanatory?
ramls/permissionOwner.json
"title" : "Owner (Module or User) that defines an Okapi Permission"

ramls/permission.json
```diff
@@ -1,58 +1,68 @@
...
-      "description": "Whether or not this permission is visible to users",
+      "description" : "Whether or not this permission is visible to users",
       "type" : "boolean"
     },
     "dummy" : {
-      "description": "Is this a temporary dummy permission",
+      "description" : "Is this a temporary dummy permission",
       "type" : "boolean"
     },
+    "inactive" : {
```
julianladisch, line 54: This is pure cosmetics but I would prefer a non-negated name like "active" or "deleted".

src/main/java/org/folio/rest/impl/PermissionUtils.java
```diff
@@ -0,0 +1,56 @@
...
+  }
+  
+  /**
+   * Compare two objects for equality, w/ checks for null, etc.
+   * 
+   * @param a LHS of comparison
+   * @param b RHS of comparison
+   * @return true if equal, false otherwise
+   */
+  public static boolean equals(Object a, Object b) {
```
julianladisch, line 19: Drop and use https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Objects.html#equals(java.lang.Object,java.lang.Object)

src/main/java/org/folio/rest/impl/PermissionUtils.java
```diff
@@ -0,0 +1,56 @@
...
+
+  /**
+   * Compare an OkapiPermission to a Permission, including definedBy.
+   * 
+   * @param okapiPerm a permission passed from Okapi
+   * @param moduleName the module name to compare against permission's definedBy.moduleName property
+   * @param perm the Permission to compare against
+   * @return true if equal, false otherwise
+   */
+  public static boolean comparePerms(OkapiPermission okapiPerm, String moduleName, Permission perm) {
```
julianladisch, line 34: The method name should be `equals`. A Java `compare` method returns an int, not a boolean.

src/main/java/org/folio/rest/impl/PermissionUtils.java
```diff
@@ -0,0 +1,56 @@
...
+   */
+  public static boolean equals(Object a, Object b) {
+    if ((a == null && b != null) || (a != null && b == null)) {
+      return false;
+    }
+    return (a == null && b == null) || (a.equals(b));
+  }
+
+  /**
+   * Compare an OkapiPermission to a Permission, including definedBy.
```
julianladisch, line 27:
```suggestion
   * Returns true if OkapiPermission and moduleName equals to the values in Permission.
```

src/main/java/org/folio/rest/impl/PermissionUtils.java
```diff
@@ -0,0 +1,56 @@
...
+   * 
+   * @param okapiPerm a permission passed from Okapi
+   * @param moduleName the module name to compare against permission's definedBy.moduleName property
+   * @param perm the Permission to compare against
+   * @return true if equal, false otherwise
+   */
+  public static boolean comparePerms(OkapiPermission okapiPerm, String moduleName, Permission perm) {
+    return comparePerms(okapiPerm, perm)
+        && perm.getDefinedBy() != null 
+        && equals(moduleName, perm.getDefinedBy().getModuleName());
```
julianladisch, line 37:
```suggestion
    String otherModuleName = perm.getDefinedBy() == null ? null : perm.getDefinedBy().getModuleName();
    return comparePerms(okapiPerm, perm)
        && Objects.equals(moduleName, otherModuleName);
```
The method should return true if perm.getDefinedBy() == null and moduleName == null.

src/main/java/org/folio/rest/impl/PermissionUtils.java
```diff
@@ -0,0 +1,56 @@
...
+  }
+  
+  /**
+   * Compare an OkapiPermission to a Permission
```
julianladisch, line 41:
```suggestion
   * Returns true if OkapiPermission equals Permission
```

src/main/java/org/folio/rest/impl/PermissionUtils.java
```diff
@@ -0,0 +1,56 @@
...
+   * 
+   * @param okapiPerm a permission passed from Okapi
+   * @param perm the Permission to compare against
+   * @return true if equal, false otherwise
+   */
+  public static boolean comparePerms(OkapiPermission okapiPerm, Permission perm) {
```
julianladisch, line 47: The method name should be `equals`. A Java `compare` method returns an int, not a boolean.

src/main/java/org/folio/rest/impl/PermissionUtils.java
```diff
@@ -0,0 +1,56 @@
...
+    if ((okapiPerm == null || perm == null)) {
+      return false;
+    }
```
julianladisch, line 50: okapi==null && perm==null should return true

src/main/java/org/folio/rest/impl/PermissionUtils.java
```diff
@@ -0,0 +1,56 @@
...
+  
+  /**
+   * Compare an OkapiPermission to a Permission
+   * 
+   * @param okapiPerm a permission passed from Okapi
+   * @param perm the Permission to compare against
+   * @return true if equal, false otherwise
+   */
+  public static boolean comparePerms(OkapiPermission okapiPerm, Permission perm) {
+    if ((okapiPerm == null || perm == null)) {
```
julianladisch, line 48:
```suggestion
    if (okapiPerm == null || perm == null) {
```

src/main/java/org/folio/rest/impl/PermsAPI.java
```diff
@@ -1080,25 +1085,33 @@ public void deletePermsPermissionsById(String id,
...
                                   int length, int start, String sortBy, String query, String memberOf,
                                   String ownedBy, Map<String, String> okapiHeaders,
                                   Handler<AsyncResult<Response>> asyncResultHandler,
                                   Context vertxContext) {
 
     try {
       boolean includeDummyPerms = "true".equals(includeDummy);
-      String[] queryArr = new String[]{""};
+      boolean includeInactivePerms = "true".equals(includeInactive);
+      String[] queryArr = new String[]{query};
```
julianladisch, line 1097: `queryArr` is not (or no longer) used in a callback.
Simply rename the query parameter to query0.
Then use
`String query = query0 == null ? "" : query0;`

src/main/java/org/folio/rest/impl/TenantPermsAPI.java
```diff
@@ -44,10 +59,10 @@ public void postTenantpermissions(OkapiPermissionSet entity, Map<String, String>
                                     Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
     try {
       String tenantId = TenantTool.tenantId(okapiHeaders);
-      savePermList(entity.getPerms(), vertxContext, tenantId).onComplete(savePermsRes -> {
-        if (savePermsRes.failed()) {
-          String err = savePermsRes.cause().getMessage();
-          logger.error(err, savePermsRes.cause());
+      handlePermLists(entity, vertxContext, tenantId).onComplete(ar -> {
```
julianladisch, line 62: Replacing onComplete by onFailure and onSucces is more concise. https://github.com/folio-org/raml-module-builder/blob/master/doc/futurisation.md#result-and-error-handling

src/main/java/org/folio/rest/impl/TenantPermsAPI.java
```diff
@@ -62,19 +77,396 @@ public void postTenantpermissions(OkapiPermissionSet entity, Map<String, String>
...
+        .onComplete(getResult -> {
+          if (getResult.failed()) {
+            Throwable cause = getResult.cause();
+            logger.error(cause.getMessage(), cause);
+            promise.fail(cause);
+            return;
+          }
+
+          List<Permission> existing = getResult.result();
+          if (existing.isEmpty() && permSet != null) {
```
julianladisch, line 95: This should be rewritten to make use of Future compose: https://github.com/folio-org/raml-module-builder/blob/master/doc/futurisation.md#result-and-error-handling
This will avoid most error handling code and the try-catch-clause and the promise.

    return getPermsByModule(moduleId, vertxContext, tenantId)
      .compose(existing -> {
        if (! existing.isEmpty() || permSet == null) {
          return Future.succeededFuture(existing);
        }
        ...
        return CompositeFuture.all(futures).map(ret);
      })
      .onFailure(e -> logger.error(e.getMessage(), e));

src/main/java/org/folio/rest/impl/TenantPermsAPI.java
```diff
@@ -62,19 +77,396 @@ public void postTenantpermissions(OkapiPermissionSet entity, Map<String, String>
...
+            // this means either:
+            // A) the first time enabling this module, or
+            // B) the permissions exist but don't yet have the definedBy field
+            List<OkapiPermission> perms = permSet.stream()
+                .filter(p -> p.getPermissionName() != null)
+                .collect(Collectors.toList());
+            List<Permission> ret = new ArrayList<>();
+
+            List<Future> futures = new ArrayList<>(perms.size());
```
julianladisch, line 104: Use proper generics with https://github.com/folio-org/okapi/blob/master/okapi-common/src/main/java/org/folio/okapi/common/GenericCompositeFuture.java

ramls/permission.json
hjiebsco, line 54: How about using `deprecated` to be consistent with what we decided for UI display?

src/main/java/org/folio/rest/impl/TenantPermsAPI.java
```diff
@@ -62,19 +77,396 @@ public void postTenantpermissions(OkapiPermissionSet entity, Map<String, String>
...
+            return;
+          }
+
+          List<Permission> existing = getResult.result();
+          if (existing.isEmpty() && permSet != null) {
+            // this means either:
+            // A) the first time enabling this module, or
+            // B) the permissions exist but don't yet have the definedBy field
+            List<OkapiPermission> perms = permSet.stream()
+                .filter(p -> p.getPermissionName() != null)
```
hjiebsco, line 100: I see this filter is used a lot. Could there be a permission without name?

src/main/java/org/folio/rest/impl/TenantPermsAPI.java
```diff
@@ -62,19 +77,396 @@ public void postTenantpermissions(OkapiPermissionSet entity, Map<String, String>
...
+        })
+        .forEach(newPerms::add);
+    }
+
+    return newPerms;
+  }
+
+  private Map<OkapiPermission, Permission> getRenamedPerms(Map<String, Permission> dbPerms,
+      List<OkapiPermission> perms) {
+    Map<OkapiPermission, Permission> renamedPerms = new HashMap<>();
```
hjiebsco, line 159: Consider to handle the case where more than one old perms are renamed/replaced by a common new perm

src/main/java/org/folio/rest/impl/TenantPermsAPI.java
```diff
@@ -62,19 +77,396 @@ public void postTenantpermissions(OkapiPermissionSet entity, Map<String, String>
...
+    List<String> permNames = new ArrayList<>();
+
+    if (perms != null) {
+      perms.stream()
+        .filter(perm -> perm.getPermissionName() != null)
+        .forEach(perm -> permNames.add(perm.getPermissionName()));
+    }
+
+    dbPerms.values().stream()
+      .filter(dbPerm -> !permNames.contains(dbPerm.getPermissionName()))
```
hjiebsco, line 215: This will include those renamed perms again. No harm but not efficient. Just a FYI

src/main/java/org/folio/rest/impl/TenantPermsAPI.java
```diff
@@ -62,19 +77,396 @@ public void postTenantpermissions(OkapiPermissionSet entity, Map<String, String>
...
+      getPermsForModule(moduleId, perms, vertxContext, tenantId).compose(existing -> {
+        Map<String, Permission> dbPerms = new HashMap<>(existing.size());
+        existing.forEach(dbPerm -> dbPerms.put(dbPerm.getPermissionName(), dbPerm));
+
+        return handleNewAndModifiedPerms(moduleId, getNewAndModifiedPerms(moduleId, dbPerms, perms),
+            vertxContext, tenantId)
+                .compose(v -> handleRenamedPerms(moduleId, getRenamedPerms(dbPerms, perms),
+                    vertxContext, tenantId))
+                .compose(v -> handleRemovedPerms(getRemovedPerms(dbPerms, perms), vertxContext,
+                    tenantId));
```
hjiebsco, line 238: This will loop through modules multiple times. Not efficient. Just a FYI.

src/main/java/org/folio/rest/impl/TenantPermsAPI.java
```diff
@@ -62,19 +77,396 @@ public void postTenantpermissions(OkapiPermissionSet entity, Map<String, String>
...
+            // A) the first time enabling this module, or
+            // B) the permissions exist but don't yet have the definedBy field
+            List<OkapiPermission> perms = permSet.stream()
+                .filter(p -> p.getPermissionName() != null)
+                .collect(Collectors.toList());
+            List<Permission> ret = new ArrayList<>();
+
+            List<Future> futures = new ArrayList<>(perms.size());
+            perms.forEach(perm -> {
+                futures.add(getModulePermByName(perm.getPermissionName(), moduleId.getProduct(),
```
hjiebsco, line 106: Should the second parameter be `moduleId.getProduct()` or `moduleId.getId()`?

hjiebsco (commented):
As discussed on Slack, it would be nice to just add `moduleName` and `moduleVersion` to perm object without the extra layer of `definedByObject`. Also, consider to remove partially implemented filtering `deprecated` permissions.

src/main/java/org/folio/rest/impl/TenantPermsAPI.java
julianladisch: Good catch.
Its description says: "The unique permission name".
It should be a required non-null property.

src/main/java/org/folio/rest/impl/TenantPermsAPI.java
craigmcnally: interesting - I have to think about this one.  It had not occurred to me.

ramls/definedByObject.json
craigmcnally: we decided to remove this definedBy property altogether.  The `moduleName` & `moduleVersion` properties will live directly under permission.

ramls/permission.json
craigmcnally: using `deprecated` as suggested below.

craigmcnally: done

src/main/java/org/folio/rest/impl/TenantPermsAPI.java
craigmcnally: I agree, much better; thanks.

src/main/java/org/folio/rest/impl/TenantPermsAPI.java
craigmcnally: Just doing it once in `handlePermLists(...)` now.  See https://issues.folio.org/browse/OKAPI-974

src/main/java/org/folio/rest/impl/TenantPermsAPI.java
craigmcnally: We can't do this without pulling in newer versions of okapi-common and vertx.  Our best bet is to "fix" this in #108

craigmcnally: @MikhailFokanov FYI

src/main/java/org/folio/rest/impl/TenantPermsAPI.java
craigmcnally: Yup, made this change in several places

src/main/java/org/folio/rest/impl/TenantPermsAPI.java
craigmcnally: addressed this, and updated the tests to exercise it.

src/main/java/org/folio/rest/impl/TenantPermsAPI.java
craigmcnally: for now we're only querying on moduleName.  The code doesn't (at this point) care if it's a downgrade/upgrade.

MikhailFokanov (approved):
Looks good.

src/main/java/org/folio/rest/impl/TenantPermsAPI.java
```diff
@@ -35,46 +44,395 @@
...
+  }
+
+  private Future<List<Permission>> getPermsForModule(ModuleId moduleId, List<OkapiPermission> perms,
+      Context vertxContext, String tenantId) {
+    return getPermsByModule(moduleId, vertxContext, tenantId)
+        .compose(existing -> {
+          if (existing.isEmpty() && perms != null) {
+            // this means either:
+            // A) the first time enabling this module, or
+            // B) the permissions exist but don't yet have the definedBy field
```
hjiebsco, line 123: The condition B) is the same as `the permissions exist but don't yet have the moduleName field`

src/main/java/org/folio/rest/impl/TenantPermsAPI.java
```diff
@@ -35,46 +44,395 @@
...
+            List<Permission> ret = new ArrayList<>();
+            List<Future> futures = new ArrayList<>(perms.size());
+            perms.forEach(perm ->
+              futures.add(getModulePermByName(perm.getPermissionName(), null, vertxContext, tenantId)
+                  .compose(dbPerm -> {
+                    if (dbPerm != null && PermissionUtils.equals(perm, dbPerm)) {
+                      // (B) we have a match, but lack definedBy
```
hjiebsco, line 130: Same as `(B) we have a match, but lack moduleName`

src/main/java/org/folio/rest/impl/TenantPermsAPI.java
```diff
@@ -250,18 +715,26 @@ public void postTenantpermissions(OkapiPermissionSet entity, Map<String, String>
             Permission foundPerm = null;
             if (!returnList.isEmpty()) {
               foundPerm = returnList.get(0);
+              if (Defined.USER.equals(foundPerm.getDefined())) {
```
hjiebsco, line 718: Would checking `foundPerm.getModuleName()` work?

ramls/permission.json
```diff
@@ -1,58 +1,82 @@
...
-    "description": {
-      "description": "A description of the permission",
-      "type": "string"
-    },
-    "tags": {
-      "description": "A list of tags for searching",
-      "type": "array",
-      "items": {
-        "type": "string"
+  "$schema" : "http://json-schema.org/draft-04/schema#",
```
julianladisch, line 2: All other .json files don't have a space before `:`

ramls/permission.json
```diff
@@ -1,58 +1,82 @@
...
-      "description": "Is this a temporary dummy permission",
+      "description" : "Is this a temporary dummy permission",
       "type" : "boolean"
     },
+    "deprecated" : {
+      "description" : "Indicates whether this permissions has been marked for deletion (soft deleted)",
+      "type" : "boolean",
+      "default" : "false"
+    },
+    "defined" : {
```
julianladisch, line 59: The name suggests a boolean value.
How about "definitionType" or "definedBy"?

src/main/java/org/folio/rest/impl/TenantPermsAPI.java
```diff
@@ -35,46 +44,395 @@
...
-          logger.error(err, savePermsRes.cause());
-          asyncResultHandler.handle(Future.succeededFuture(
-              PostTenantpermissionsResponse.respond400WithTextPlain(err)));
-          return;
-        }
-        asyncResultHandler.handle(Future.succeededFuture(
-            PostTenantpermissionsResponse.respond201WithApplicationJson(entity)));
-      });
+      handlePermLists(entity, vertxContext, tenantId)
+          .compose(v -> addMissingDefinedContext(vertxContext, tenantId))
```
hjiebsco, line 61: How does this work? After handling perms for **a specific module**, then update empty `definedBy` field to be `User` for all other perms (other modules') for that tenant? During bootstrap, if tenant has 3 modules enabled. Okapi sends modA perms over first, would this incorrectly mark perms from modB and modB as `User`?

src/main/java/org/folio/rest/impl/TenantPermsAPI.java
```diff
@@ -35,46 +44,395 @@
...
+              return;
+            }
+            pgClient.endTx(connection, done -> {
+              logger.info(String.format("Added defined context to %s user-defined permissions",
+                  userDefined.size()));
+              promise.complete();
+            });
+          });
+        }));
+    return promise.future();
```
julianladisch, line 113: Is there a reason why this doesn't use a plain SQL command with PostgresClient.execute?

    UPDATE [schema].[table] SET jsonb = jsonb || '{"defined": "User"}'::jsonb WHERE jsonb->'defined' IS NULL;

src/main/java/org/folio/rest/impl/PermissionUtils.java
```diff
@@ -0,0 +1,46 @@
...
+  private PermissionUtils() {
+    
+  }
+
+  /**
+   * Returns true if the provided OkapiPermission and moduleName are equal to the values in the
+   * provided Permission.
+   * 
+   * @param okapiPerm a permission passed from Okapi
+   * @param moduleName the module name to compare against permission's definedBy.moduleName property
```
julianladisch, line 18: `definedBy.moduleName` no longer exists

src/main/java/org/folio/rest/impl/TenantPermsAPI.java
julianladisch, line 123: the `definedBy` field no longer exists

src/main/java/org/folio/rest/impl/TenantPermsAPI.java
julianladisch, line 130: `definedBy` no longer exists

src/main/java/org/folio/rest/impl/TenantPermsAPI.java
craigmcnally: <face-palm> you're right - this was pretty short-sighted on my part.  I have to work thought some examples.

ramls/permission.json
```diff
@@ -1,58 +1,82 @@
...
+    "moduleName" : {
+      "description" : "The name of the module (not including version) that defined this permission",
+      "type" : "string",
+      "readOnly" : true
+    },
+    "moduleVersion" : {
+      "description" : "The version of the module that defined this permission",
+      "type" : "string",
+      "readOnly" : true
+    },
```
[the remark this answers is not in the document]
craigmcnally: That's not my recollection.  I thought we decided to keep defined(By) and flatten the three props (move them up to Permission level), accepting that there was some overlap between `mutable` and `defined`, but since we couldn't rely on mutable as a reliable indicator that a module was defined by the system (because it wasn't locked down as read-only), this was necessary.

ramls/permission.json
craigmcnally: right, it should be definedBy.  I'll fix it.

src/main/java/org/folio/rest/impl/TenantPermsAPI.java
craigmcnally: This is the sticky point... if we do that there's no way to determine if it's 
A) a system defined permission lacking the defined context; e.g. during bootstrap, 
or 
B) a user-defined permission.

src/main/java/org/folio/rest/impl/TenantPermsAPI.java
craigmcnally: Removing this method so this is a moot point.  As @hjiebsco  points out we can't do this anyway.

src/main/java/org/folio/rest/impl/TenantPermsAPI.java
hjiebsco: > This is the sticky point... if we do that there's no way to determine if it's
> A) a system defined permission lacking the defined context; e.g. during bootstrap,
> or
> B) a user-defined permission.

Bootstrap is a special case and it has to make some assumptions. For example, if tenant has two modules (A and B) enabled, when bootstrapping the perms for the first module (mod-A), you would not know if mod-A perm provided by Okapi is really a perf defined by mod-A, mod-B or even user by looking at database. But once the bootstrap is done for all modules, we can safely to use moduleName to know whether a permission is system or user defined.

craigmcnally: Right - above where we identify that the permission provided matches one in the database, except for the moduleName I update the permission right then and there, so that later on (here) we it's no longer ambiguous.

ramls/permission.json
craigmcnally: ok fixed

ramls/permission.json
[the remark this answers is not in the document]
craigmcnally: defined/definedBy was removed.  Now Permission just has moduleName and moduleVersion.

src/main/java/org/folio/rest/impl/TenantPermsAPI.java
craigmcnally: Removed this.  See related comment threads for additional details.

src/main/java/org/folio/rest/impl/PermissionUtils.java
craigmcnally: yup, fixed

src/test/java/org/folio/permstest/RestVerticleTest.java
```diff
@@ -47,7 +54,7 @@
 import org.junit.Test;
 import org.junit.rules.Timeout;
 import org.junit.runner.RunWith;
-
+import com.sun.el.stream.Stream;
```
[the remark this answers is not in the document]
craigmcnally: removed & organized imports

src/main/java/org/folio/rest/impl/TenantPermsAPI.java
```diff
@@ -35,46 +42,376 @@
...
+          if (existing.isEmpty() && perms != null) {
+            // this means either:
+            // A) the first time enabling this module, or
+            // B) the permissions exist but don't yet have the moduleName field
+            List<Permission> ret = new ArrayList<>();
+            List<Future> futures = new ArrayList<>(perms.size());
+            perms.forEach(perm ->
+              futures.add(getModulePermByName(perm.getPermissionName(), null, vertxContext, tenantId)
+                  .compose(dbPerm -> {
+                    if (dbPerm != null && PermissionUtils.equals(perm, dbPerm)) {
```
hjiebsco, line 106: I do not think `PermissionUtils.equals(perm, dbPerm)` is needed. For example, the perm name was same but description was updated. In this case, we have to treat it as the same perm during bootstrap.

craigmcnally: When I say "bootstrap" I'm referring to a very specific thing.  In OKAPI-947 we made an adjustment to call _tenantPermissions for each module enabled for the tenant whenever mod-permissions is installed/upgraded.  During this, it's possible to accurately identify a permission which was system-defined but lacks the module context because those permissions are identical except for the absence of the module context.  That's what this test does.

craigmcnally: > For example, the perm name was same but description was updated. In this case, we have to treat it as the same perm during bootstrap
This is detected later and handled as an updated permission

hjiebsco: > When I say "bootstrap" I'm referring to a very specific thing. In OKAPI-947 we made an adjustment to call _tenantPermissions for each module enabled for the tenant whenever mod-permissions is installed/upgraded. During this, it's possible to accurately identify a permission which was system-defined but lacks the module context because those permissions are identical except for the absence of the module context. That's what this test does.

I see. You might want to add another condition to make sure `dbPerm` has no `moduleName` to cover an edge case for `A) the first time enabling this module`. For example, if two modules define the same perm (very unlikely though). Of course, if `dbPerm` has `moduleName`, you might have to throw an error because it would be a collision.

craigmcnally: yes that's a good suggestion.  I'll work that in, thanks.

adamdickmeiss (commented):
The use of error messages is not consistent.  Sometimes getMessage, sometimes getLocalizedMessage.

craigmcnally:
> The use of error messages is not consistent. Sometimes getMessage, sometimes getLocalizedMessage.

which is preferred?

adamdickmeiss:
> > The use of error messages is not consistent. Sometimes getMessage, sometimes getLocalizedMessage.
> 
> which is preferred
I prefer getMessage as this might lead to less surprises in unit tests.

adamdickmeiss:
We can also postpone till log4j2 migration which will touch ssme æines.

craigmcnally:
> We can also postpone till log4j2 migration which will touch sane lines. Lets close this

OK, makes sense.

src/main/java/org/folio/rest/impl/TenantPermsAPI.java
craigmcnally: OK - got this in w/ test case.  Note that we also have to check that the permission isn't a dummy.

craigmcnally:
I think I set a record for the longest PR...  What do I win?

src/main/java/org/folio/rest/impl/TenantPermsAPI.java
hjiebsco: Why check for dummy? Would not dummy perm have moduleName as well?

src/main/java/org/folio/rest/impl/TenantPermsAPI.java
```diff
@@ -35,46 +42,386 @@
...
+                      logger.error(msg);
+                      return Future.failedFuture(msg);
+                    }
+                    if (dbPerm != null && PermissionUtils.equals(perm, dbPerm)) {
+                      // (B) we have a match, but lack moduleName. Fix it before we add it.
+                      return addMissingModuleContext(dbPerm, moduleId, vertxContext, tenantId)
+                          .onFailure(Future::failedFuture)
+                          .onSuccess(ret::add);
+                    }
+                    return Future.succeededFuture();
```
hjiebsco, line 122: Would code block below do what we want?
```
                    if (dbPerm == null) {
                      return Future.succeededFuture();
                    }
                    if (dbPerm.getModuleName() == null && !dbPerm.getMutable() 
                        && PermissionUtils.equals(perm, dbPerm)) {
                      // (B) we have a match, but lack moduleName. Fix it before we add it.
                      return addMissingModuleContext(dbPerm, moduleId, vertxContext, tenantId)
                          .onFailure(Future::failedFuture)
                          .onSuccess(ret::add);
                    } else {
                      String msg = String.format(
                          "Collision! A Permission with name %s is already defined by %s",
                          perm.getPermissionName(), Json.encode(dbPerm));
                      logger.error(msg);
                      return Future.failedFuture(msg);
                    }
```

craigmcnally: No.  Like I said earlier we need to check dummy. Otherwise we break how dummy permissions are used as placeholders for subPermissions defined by other modules until the real thing is defined.  This is exercised in the unit tests.  See RestVerticleTest, specifically relevant are tests involving "sendOtherPermissionSet" and "sendAlienPermissionSet".

Also, sonarqube will complain about not testing mutable against the primitive on that mutable check, hence the uglier (but safer) syntax is preferred.  Possible NPE otherwise.

I will add the bit about case B matching on mutable too though.

hjiebsco: I did not mean to fix your syntax. I was just trying to help to get the logic right. For example, current implementation does not handle the case where the found duplicated perm name is a user defined perm, or dummy perm. BTW dummy perm is just the precursor of normal perm.

craigmcnally: Ah, got it @hjiebsco.  I was looking at this/responding on my phone last night and it appeared that this was a code suggestion to be applied/committed as-is.  

> For example, current implementation does not handle the cases where the found duplicated perm name is a user defined perm, or dummy perm

The cases we miss here are caught later in the processing, but it's probably best to fail early.  

> dummy perm is just the precursor of normal perm

Exactly.  That's what I was trying to say, which is why we can't consider it a collision if the found permission is a dummy.  

I've taken most your suggestion, made the necessary syntax adjustments and also continue w/o failing if the found permission is a dummy.  I think this covers all the bases.

hjiebsco: > Ah, got it @hjiebsco. I was looking at this/responding on my phone last night and it appeared that this was a code suggestion to be applied/committed as-is.
> 
> > For example, current implementation does not handle the cases where the found duplicated perm name is a user defined perm, or dummy perm
> 
> The cases we miss here are caught later in the processing, but it's probably best to fail early.
> 
> > dummy perm is just the precursor of normal perm
> 
> Exactly. That's what I was trying to say, which is why we can't consider it a collision if the found permission is a dummy.
> 
> I've taken most your suggestion, made the necessary syntax adjustments and also continue w/o failing if the found permission is a dummy. I think this covers all the bases.

About the dummy perm, for example if mod-A uses a perm but did not define it, it will become a dummy leftover. BTW mod-perm should not allow that but that's another defect for another day. Later when mod-B comes in and it also defines the same perm. This is a collision in my view, isn't it?

craigmcnally: Not in my mind.  If it's a dummy permission it's just a placeholder and can safely be replaced with a non-dummy permission.

hjiebsco: Dummy is just a concept in mod-permissions' world. From developers' perspective, they have no idea about whether their perms will be tagged as dummy or not. In any case, duplicated permission names between modules is a collision, isn't? Anyway, this would not be a problem if mod-perms does not have that pre-existing defect.

src/main/java/org/folio/rest/impl/PermissionUtils.java
```diff
@@ -0,0 +1,46 @@
...
+   * @param perm the Permission to compare against
+   * @return true if equal, false otherwise
+   */
+  public static boolean equals(OkapiPermission okapiPerm, String moduleName, Permission perm) {
+    String otherModuleName = perm == null ? null : perm.getModuleName();
+    return equals(okapiPerm, perm) && Objects.equals(moduleName, otherModuleName);
+  }
+  
+  /**
+   * Returns true if the provide OkapiPermission is equal to the values in provided Permission.
```
julianladisch, line 28:
```suggestion
   * Returns true if the values in okapiPerm equal to the values in perm, or if both okapiPerm and perm are null.
```

src/main/java/org/folio/rest/impl/TenantPermsAPI.java
craigmcnally: We discussed this off-line and decided to take this to the rest of the team for discussion and possibly follow-on work.  We agreed it shouldn't block this PR from being merged.

hjiebsco (approved):
Looks good. Note, the dummy perm is a stick point and we will handle that after team discussion tomorrow.

PR #109, opened 2021-02-01
Provide _tenant v2 - fixes MODPERMS-128

adamdickmeiss:


funkymalc (approved):

wafschneider (approved):
Module descriptor update looks correct. I can only assume the code to handle it is there.

PR #110, opened 2021-02-06
Allow re-use of deprecated permissions MODPERMS-130

adamdickmeiss:
This PR fixes the problems as described in MODPERMS-130.

It also fixes a problem where permission could get a new module context if the structure was the same. That means multiple modules could be enabled and definitions could start to differ for same permission.

### DIFF
--- a/src/main/java/org/folio/rest/impl/TenantPermsAPI.java
+++ b/src/main/java/org/folio/rest/impl/TenantPermsAPI.java
@@ -106,8 +106,8 @@ perm.setModuleName(moduleId.getProduct());
                     if (dbPerm == null || Boolean.TRUE.equals(dbPerm.getDummy())) {
                       return Future.succeededFuture();
                     }
-                    if (dbPerm != null && Boolean.FALSE.equals(dbPerm.getMutable())
-                        && PermissionUtils.equals(perm, dbPerm)) {
+                    if (Boolean.TRUE.equals(dbPerm.getDeprecated()) || (Boolean.FALSE.equals(dbPerm.getMutable())
+                        && PermissionUtils.equals(perm, dbPerm))) {
                       // (B) we have a match, but lack moduleName. Fix it before we add it.
                       return addMissingModuleContext(dbPerm, moduleId, vertxContext, tenantId)
                           .onFailure(Future::failedFuture)
--- a/src/test/java/org/folio/permstest/RestVerticleTest.java
+++ b/src/test/java/org/folio/permstest/RestVerticleTest.java
@@ -328,6 +328,41 @@ public class RestVerticleTest {
     context.assertEquals(201, response.code);
   }
 
+  // Test that a permission can be used in another module if first one is deleted MODPERMS-130
+  @Test
+  public void testPermissionsOnTheMove(TestContext context) {
+    String perm = "perm" + UUID.randomUUID().toString();
+    JsonObject permissionSet = new JsonObject()
+        .put("moduleId","moduleA")
+        .put("perms", new JsonArray()
+            .add(new JsonObject()
+                .put("permissionName", perm)
+                .put("displayName", "Description A")
+                )
+            );
+    Response response = send(HttpMethod.POST, "/_/tenantpermissions", permissionSet.encode(), context);
+    context.assertEquals(201, response.code);
+
+    // remove permissions for moduleA
+    permissionSet = new JsonObject()
+        .put("moduleId","moduleA")
+        .put("perms", new JsonArray());
+    response = send(HttpMethod.POST, "/_/tenantpermissions", permissionSet.encode(), context);
+    context.assertEquals(201, response.code);
+
+    // use same permission again in other module with new definition
+    permissionSet = new JsonObject()
+        .put("moduleId","moduleB")
+        .put("perms", new JsonArray()
+            .add(new JsonObject()
+                .put("permissionName", perm)
+                .put("displayName", "Description B")
+            )
+        );
+    response = send(HttpMethod.POST, "/_/tenantpermissions", permissionSet.encode(), context);
+    context.assertEquals(201, response.code);
+  }
+
   @Test
   public void testPutPermsUsersByIdDummyPerm(TestContext context) {
     String postPermUsersRequest = "{\"userId\": \""+ userUserId +"\",\"permissions\": " +
--- a/src/test/java/org/folio/permstest/RestVerticleTest.java
+++ b/src/test/java/org/folio/permstest/RestVerticleTest.java
@@ -364,6 +364,57 @@ public class RestVerticleTest {
   }
 
   @Test
+  public void testPermissionsChangingInMultipleModules(TestContext context) {
+    String perm = "perm" + UUID.randomUUID().toString();
+    JsonObject permissionSet = new JsonObject()
+        .put("moduleId","moduleA-1.0.0")
+        .put("perms", new JsonArray()
+            .add(new JsonObject()
+                .put("permissionName", perm)
+                .put("displayName", "Description 1")
+            )
+        );
+    Response response = send(HttpMethod.POST, "/_/tenantpermissions", permissionSet.encode(), context);
+    context.assertEquals(201, response.code);
+
+    // use same permission in other module with same definition
+    permissionSet = new JsonObject()
+        .put("moduleId","moduleB-1.0.0")
+        .put("perms", new JsonArray()
+            .add(new JsonObject()
+                .put("permissionName", perm)
+                .put("displayName", "Description 1")
+            )
+        );
+    response = send(HttpMethod.POST, "/_/tenantpermissions", permissionSet.encode(), context);
+    context.assertEquals(201, response.code);
+
+    // lets change the definition in other module
+    permissionSet = new JsonObject()
+        .put("moduleId","moduleB-1.0.1")
+        .put("perms", new JsonArray()
+            .add(new JsonObject()
+                .put("permissionName", perm)
+                .put("displayName", "Description 2")
+            )
+        );
+    response = send(HttpMethod.POST, "/_/tenantpermissions", permissionSet.encode(), context);
+    context.assertEquals(201, response.code);
+
+    // module A upgraded with its ORIGINAL definition gives collision!
+    permissionSet = new JsonObject()
+        .put("moduleId","moduleA-1.0.1")
+        .put("perms", new JsonArray()
+            .add(new JsonObject()
+                .put("permissionName", perm)
+                .put("displayName", "Description 1")
+            )
+        );
+    response = send(HttpMethod.POST, "/_/tenantpermissions", permissionSet.encode(), context);
+    context.assertEquals(400, response.code);
+  }
+
+  @Test
   public void testPutPermsUsersByIdDummyPerm(TestContext context) {
     String postPermUsersRequest = "{\"userId\": \""+ userUserId +"\",\"permissions\": " +
         "[], \"id\" : \"" + userId2 + "\"}";


src/main/java/org/folio/rest/impl/TenantPermsAPI.java
```diff
@@ -106,8 +106,8 @@ public void postTenantpermissions(OkapiPermissionSet entity, Map<String, String>
                     if (dbPerm == null || Boolean.TRUE.equals(dbPerm.getDummy())) {
                       return Future.succeededFuture();
                     }
-                    if (dbPerm != null && Boolean.FALSE.equals(dbPerm.getMutable())
-                        && PermissionUtils.equals(perm, dbPerm)) {
+                    if (Boolean.TRUE.equals(dbPerm.getDeprecated()) || (Boolean.FALSE.equals(dbPerm.getMutable())
+                        && PermissionUtils.equals(perm, dbPerm))) {
```
craigmcnally, line 110: Can you please update the comment below?

hjiebsco, line 110: Note `PermissionUtils.equals(perm, dbPerm)` will compare perm display name as well. What's the chance that the new perm will also have `(Deprecated)` as prefix?

hjiebsco (commented):
Would module downgrade still be supported if the context is overwritten by another module during reuse?

src/main/java/org/folio/rest/impl/TenantPermsAPI.java
skoczko: @adamdickmeiss @craigmcnally Guys, shouldn't all comparison be done with the permission name (not displayName)?

adamdickmeiss: Sure.

adamdickmeiss: @hjiebsco .. Doesn't matter.. It will fail regardless of display name and my test shows it.  This condition should *not* use PermissionUtils.equals as this is not relevant. Either there's a colliision because a permission is defined elsewhere or NOT.

It will be utterly confusing if we allowed same name in two modules and .. suddenly because you change the definition in one of the modules there *might* be a conflict. I say *might* because it might be depending on the order of insertion because the context only allows ONE module to be attached to a permission.

adamdickmeiss: The permissionName is already the same.. The check can be simplified to:
```java
  if (Boolean.TRUE.equals(dbPerm.getDeprecated()) || Boolean.FALSE.equals(dbPerm.getMutable())) {
```

hjiebsco: I will mark this as resolved. But we still need to decide the implication of re-use to downgrade.

hjiebsco (commented):
If all modules define permission name based on certain naming convention, maybe we can avoid naming conflict / reuse between modules?

adamdickmeiss:
> Would module downgrade still be supported if the context is overwritten by another module during reuse?

Upgrade/downgrade is not a concern here AFAIK. If a module transitions from one version to another the new version's permissions are posted.. Any permissions that is no longer a member of that module will be deprecated.. The context is allowed to be changed when it's deprecated.

adamdickmeiss:
> If all modules define permission name based on certain naming convention, maybe we can avoid naming conflict / reuse between modules?

Yes. It would have been nice to avoid these duplicates with some scheme or automatic way. Like interfaces which are also names, but there has been a restriction on that from day one.

hjiebsco:
> > Would module downgrade still be supported if the context is overwritten by another module during reuse?
> 
> Upgrade/downgrade is not a concern here AFAIK. If a module transitions from one version to another the new version's permissions are posted.. Any permissions that is no longer a member of that module will be deprecated.. The context is allowed to be changed when it's deprecated.

Let's use an example: modA has permA defined in v1 and userA has that permA assigned. Next, modA is upgraded to v2 and the permA in v1 is marked as deprecated but userA still has that permA assigned (soft delete as what has been decided and implemented). Then modB comes in and reused permA, what does that mean to the userA permA assignment? If you remove that assignment, what if modA needs to downgrade from v2 to v1? The userA will not have that permA assignment anymore.

adamdickmeiss:
> > > Would module downgrade still be supported if the context is overwritten by another module during reuse?
> > 
> > 
> > Upgrade/downgrade is not a concern here AFAIK. If a module transitions from one version to another the new version's permissions are posted.. Any permissions that is no longer a member of that module will be deprecated.. The context is allowed to be changed when it's deprecated.
> 
> Let's use an example: modA has permA defined in v1 and userA has that permA assigned. Next, modA is upgraded to v2 and the permA in v1 is marked as deprecated but userA still has that permA assigned (soft delete as what has been decided and implemented). Then modB comes in and reused permA, what does that mean to the userA permA assignment? If you remove that assignment, what if modA needs to downgrade from v2 to v1? The userA will not have that permA assignment anymore.
u



> > > Would module downgrade still be supported if the context is overwritten by another module during reuse?
> > 
> > 
> > Upgrade/downgrade is not a concern here AFAIK. If a module transitions from one version to another the new version's permissions are posted.. Any permissions that is no longer a member of that module will be deprecated.. The context is allowed to be changed when it's deprecated.
> 
> Let's use an example: modA has permA defined in v1 and userA has that permA assigned. Next, modA is upgraded to v2 and the permA in v1 is marked as deprecated but userA still has that permA assigned (soft delete as what has been decided and implemented). Then modB comes in and reused permA, what does that mean to the userA permA assignment? If you remove that assignment, what if modA needs to downgrade from v2 to v1? The userA will not have that permA assignment anymore.

userA will keep permissions I hope. So she should have permission to use modB , then later (back to) modA.

hjiebsco:
> > > > Would module downgrade still be supported if the context is overwritten by another module during reuse?
> > > 
> > > 
> > > Upgrade/downgrade is not a concern here AFAIK. If a module transitions from one version to another the new version's permissions are posted.. Any permissions that is no longer a member of that module will be deprecated.. The context is allowed to be changed when it's deprecated.
> > 
> > 
> > Let's use an example: modA has permA defined in v1 and userA has that permA assigned. Next, modA is upgraded to v2 and the permA in v1 is marked as deprecated but userA still has that permA assigned (soft delete as what has been decided and implemented). Then modB comes in and reused permA, what does that mean to the userA permA assignment? If you remove that assignment, what if modA needs to downgrade from v2 to v1? The userA will not have that permA assignment anymore.
> > u
> 
> > > > Would module downgrade still be supported if the context is overwritten by another module during reuse?
> > > 
> > > 
> > > Upgrade/downgrade is not a concern here AFAIK. If a module transitions from one version to another the new version's permissions are posted.. Any permissions that is no longer a member of that module will be deprecated.. The context is allowed to be changed when it's deprecated.
> > 
> > 
> > Let's use an example: modA has permA defined in v1 and userA has that permA assigned. Next, modA is upgraded to v2 and the permA in v1 is marked as deprecated but userA still has that permA assigned (soft delete as what has been decided and implemented). Then modB comes in and reused permA, what does that mean to the userA permA assignment? If you remove that assignment, what if modA needs to downgrade from v2 to v1? The userA will not have that permA assignment anymore.
> 
> userA will keep permissions I hope. So she should have permission to use modB , then later (back to) modA.

But how can we know that userA should have the reused permA for modB?

adamdickmeiss:
> > > > > Would module downgrade still be supported if the context is overwritten by another module during reuse?
> > > > 
> > > > 
> > > > Upgrade/downgrade is not a concern here AFAIK. If a module transitions from one version to another the new version's permissions are posted.. Any permissions that is no longer a member of that module will be deprecated.. The context is allowed to be changed when it's deprecated.
> > > 
> > > 
> > > Let's use an example: modA has permA defined in v1 and userA has that permA assigned. Next, modA is upgraded to v2 and the permA in v1 is marked as deprecated but userA still has that permA assigned (soft delete as what has been decided and implemented). Then modB comes in and reused permA, what does that mean to the userA permA assignment? If you remove that assignment, what if modA needs to downgrade from v2 to v1? The userA will not have that permA assignment anymore.
> > > u
> > 
> > 
> > > > > Would module downgrade still be supported if the context is overwritten by another module during reuse?
> > > > 
> > > > 
> > > > Upgrade/downgrade is not a concern here AFAIK. If a module transitions from one version to another the new version's permissions are posted.. Any permissions that is no longer a member of that module will be deprecated.. The context is allowed to be changed when it's deprecated.
> > > 
> > > 
> > > Let's use an example: modA has permA defined in v1 and userA has that permA assigned. Next, modA is upgraded to v2 and the permA in v1 is marked as deprecated but userA still has that permA assigned (soft delete as what has been decided and implemented). Then modB comes in and reused permA, what does that mean to the userA permA assignment? If you remove that assignment, what if modA needs to downgrade from v2 to v1? The userA will not have that permA assignment anymore.
> > 
> > 
> > userA will keep permissions I hope. So she should have permission to use modB , then later (back to) modA.
> 
> But how can we know that userA should have the reused permA for modB?

When a permission is marked deprecated (and that happens when it's not part of a module definition anymore) and the context is correct.. Then if the system ALSO removed that permission for that user, then obviously that information is lost that a user had that permission.

Which is why I said I hope the user still has the permission for it (even though it's deprecated). That means, if another module (including the same module where it was originally defined) defines it again, then the user is not affected. And so downgrade would just work. I thought the idea of deprecated permissions was a soft delete. If at any time, two modules start to define the permission, then there should be a conflict.

Example.. A module defines a set of permissions.. But there's another module defiing the same permissionNames (say another implementation of mod-authtoken or mod-login-saml or something).. then we should be able to go from one to another.. One will be disabled and another one will be disabled.. It's not a something that is obscure .. It may happen.

hjiebsco:
> > > > > > Would module downgrade still be supported if the context is overwritten by another module during reuse?
> > > > > 
> > > > > 
> > > > > Upgrade/downgrade is not a concern here AFAIK. If a module transitions from one version to another the new version's permissions are posted.. Any permissions that is no longer a member of that module will be deprecated.. The context is allowed to be changed when it's deprecated.
> > > > 
> > > > 
> > > > Let's use an example: modA has permA defined in v1 and userA has that permA assigned. Next, modA is upgraded to v2 and the permA in v1 is marked as deprecated but userA still has that permA assigned (soft delete as what has been decided and implemented). Then modB comes in and reused permA, what does that mean to the userA permA assignment? If you remove that assignment, what if modA needs to downgrade from v2 to v1? The userA will not have that permA assignment anymore.
> > > > u
> > > 
> > > 
> > > > > > Would module downgrade still be supported if the context is overwritten by another module during reuse?
> > > > > 
> > > > > 
> > > > > Upgrade/downgrade is not a concern here AFAIK. If a module transitions from one version to another the new version's permissions are posted.. Any permissions that is no longer a member of that module will be deprecated.. The context is allowed to be changed when it's deprecated.
> > > > 
> > > > 
> > > > Let's use an example: modA has permA defined in v1 and userA has that permA assigned. Next, modA is upgraded to v2 and the permA in v1 is marked as deprecated but userA still has that permA assigned (soft delete as what has been decided and implemented). Then modB comes in and reused permA, what does that mean to the userA permA assignment? If you remove that assignment, what if modA needs to downgrade from v2 to v1? The userA will not have that permA assignment anymore.
> > > 
> > > 
> > > userA will keep permissions I hope. So she should have permission to use modB , then later (back to) modA.
> > 
> > 
> > But how can we know that userA should have the reused permA for modB?
> 
> When a permission is marked deprecated (and that happens when it's not part of a module definition anymore) and the context is correct.. Then if the system ALSO removed that permission for that user, then obviously that information is lost that a user had that permission.
> 
> Which is why I said I hope the user still has the permission for it (even though it's deprecated). That means, if another module (including the same module where it was originally defined) defines it again, then the user is not affected. And so downgrade would just work. I thought the idea of deprecated permissions was a soft delete. If at any time, two modules start to define the permission, then there should be a conflict.
> 
> Example.. A module defines a set of permissions.. But there's another module defiing the same permissionNames (say another implementation of mod-authtoken or mod-login-saml or something).. then we should be able to go from one to another.. One will be disabled and another one will be disabled.. It's not a something that is obscure .. It may happen.
> 
> I thought we'd have a special call or A UI to REALLY remove the deprecated permissions.. Perhaps it's done for all. Perhaps it's done individually.. But by default, it's a soft delete.
> 
> I probably misunderstood what meant @hjiebsco :-)

If the same permission is redefined by another implementation of the same interface, re-use makes sense because it is essentially the same permission for the same purpose. But if it is redefined by another unrelated module for a different purpose, the re-use will create a complicated scenario.

adamdickmeiss:
> > > > > > > Would module downgrade still be supported if the context is overwritten by another module during reuse?
> > > > > > 
> > > > > > 
> > > > > > Upgrade/downgrade is not a concern here AFAIK. If a module transitions from one version to another the new version's permissions are posted.. Any permissions that is no longer a member of that module will be deprecated.. The context is allowed to be changed when it's deprecated.
> > > > > 
> > > > > 
> > > > > Let's use an example: modA has permA defined in v1 and userA has that permA assigned. Next, modA is upgraded to v2 and the permA in v1 is marked as deprecated but userA still has that permA assigned (soft delete as what has been decided and implemented). Then modB comes in and reused permA, what does that mean to the userA permA assignment? If you remove that assignment, what if modA needs to downgrade from v2 to v1? The userA will not have that permA assignment anymore.
> > > > > u
> > > > 
> > > > 
> > > > > > > Would module downgrade still be supported if the context is overwritten by another module during reuse?
> > > > > > 
> > > > > > 
> > > > > > Upgrade/downgrade is not a concern here AFAIK. If a module transitions from one version to another the new version's permissions are posted.. Any permissions that is no longer a member of that module will be deprecated.. The context is allowed to be changed when it's deprecated.
> > > > > 
> > > > > 
> > > > > Let's use an example: modA has permA defined in v1 and userA has that permA assigned. Next, modA is upgraded to v2 and the permA in v1 is marked as deprecated but userA still has that permA assigned (soft delete as what has been decided and implemented). Then modB comes in and reused permA, what does that mean to the userA permA assignment? If you remove that assignment, what if modA needs to downgrade from v2 to v1? The userA will not have that permA assignment anymore.
> > > > 
> > > > 
> > > > userA will keep permissions I hope. So she should have permission to use modB , then later (back to) modA.
> > > 
> > > 
> > > But how can we know that userA should have the reused permA for modB?
> > 
> > 
> > When a permission is marked deprecated (and that happens when it's not part of a module definition anymore) and the context is correct.. Then if the system ALSO removed that permission for that user, then obviously that information is lost that a user had that permission.
> > Which is why I said I hope the user still has the permission for it (even though it's deprecated). That means, if another module (including the same module where it was originally defined) defines it again, then the user is not affected. And so downgrade would just work. I thought the idea of deprecated permissions was a soft delete. If at any time, two modules start to define the permission, then there should be a conflict.
> > Example.. A module defines a set of permissions.. But there's another module defiing the same permissionNames (say another implementation of mod-authtoken or mod-login-saml or something).. then we should be able to go from one to another.. One will be disabled and another one will be disabled.. It's not a something that is obscure .. It may happen.
> > I thought we'd have a special call or A UI to REALLY remove the deprecated permissions.. Perhaps it's done for all. Perhaps it's done individually.. But by default, it's a soft delete.
> > I probably misunderstood what meant @hjiebsco :-)
> 
> If the same permission is redefined by another implementation of the same interface, re-use makes sense because it is essentially the same permission for the same purpose. But if it is redefined by another unrelated module for a different purpose, the re-use will create a complicated scenario.

That could happen that somebody hijacked a permission. But so far this has never been the case.. When we have seen permissions being re-used they were for the same purpose. Again .. in order for this to happen one would have to bet on the other one being gone.

adamdickmeiss:
Closing as GitHub/GitHub seems to be stock.. Continued in https://github.com/folio-org/mod-permissions/pull/112

PR #111, opened 2021-02-06
Fix minor issues reported by IntelliJ IDEA MODPERMS-131

adamdickmeiss:
Should we import okapi-common and also get rid of untyped Future(s), Promise(s)?

### DIFF
--- a/src/main/java/org/folio/rest/impl/PermsAPI.java
+++ b/src/main/java/org/folio/rest/impl/PermsAPI.java
@@ -62,10 +62,10 @@ public class PermsAPI implements Perms {
 
   public static class FieldUpdateValues {
 
-    private String fieldValue;
-    private String permissionName;
-    private PermissionField field;
-    private Operation operation;
+    private final String fieldValue;
+    private final String permissionName;
+    private final PermissionField field;
+    private final Operation operation;
 
     public FieldUpdateValues(String fieldValue, String permissionName,
                              PermissionField field, Operation operation) {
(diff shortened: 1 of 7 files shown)


craigmcnally:
Yes I think we were waiting until we upgraded to a newer version of rmb.

src/test/java/org/folio/permstest/TestUtil.java
```diff
@@ -72,8 +59,8 @@ public JsonObject getJson() {
       .putHeader("content-type", CONTENT_TYPE_JSON)
       .putHeader("accept", CONTENT_TYPE_JSON);
     if (headers != null) {
-      for (Map.Entry entry : headers.entries()) {
-        request.putHeader((String) entry.getKey(), (String) entry.getValue());
+      for (Map.Entry<String, String> entry : headers.entries()) {
+        request.putHeader(entry.getKey(), entry.getValue());
```
julianladisch, line 63: Can you use HttpRequest#putHeaders(MultiMap) ?
https://github.com/vert-x3/vertx-web/blob/4.0.0/vertx-web-client/src/main/java/io/vertx/ext/web/client/impl/HttpRequestImpl.java#L160-L164

src/main/java/org/folio/rest/impl/PermsAPI.java
```diff
@@ -193,31 +193,30 @@ void postPermsUsersTrans(PermissionUser entity, Context vertxContext,
...
+                                  ID_FIELD, permUser.getId(),
+                                  UNABLE_TO_UPDATE_DERIVED_FIELDS + cause.getMessage()))));
+                    } else {
+                      asyncResultHandler.handle(Future.succeededFuture(
+                          PostPermsUsersResponse.respond500WithTextPlain(
+                              cause.getMessage())));
+                    }
+                  })
+              )
+              .onSuccess(res1 ->
```
hjiebsco, line 214: Maybe `res` not `res1`?

src/test/java/org/folio/rest/impl/PermsAPITest.java
```diff
@@ -95,7 +95,7 @@ public void testUpdateUserPermissionsFail(TestContext context) {
     Context vertxContext = vertx.getOrCreateContext();
     PostgresClient postgresClient = PostgresClient.getInstance(vertxContext.owner(), tenantId);
     postgresClient.startTx(s -> {
-      Future<Void> future = PermsAPI.updateUserPermissions(s, "bad",
+      Future<Void> future = api.updateUserPermissions(s, "bad",
```
hjiebsco, line 98: `updateUserPermissions` is a static method. Consider to use `PermsAPI.updateUserPermissions` but remove the unnecessary line 92 `PermsAPI api = new PermsAPI();`

src/main/java/org/folio/rest/impl/PermsAPI.java
adamdickmeiss: yep

src/test/java/org/folio/permstest/TestUtil.java
adamdickmeiss: yes,

src/test/java/org/folio/rest/impl/PermsAPITest.java
adamdickmeiss: ok

hjiebsco (approved):

PR #112, opened 2021-02-09
Allow reuse of permissions MODPERMS-130

adamdickmeiss:
This PR fixes the problems as described in MODPERMS-130.

It also fixes a problem where permission could get a new module context if the structure was the same. That means multiple modules could be enabled and definitions could start to differ for same permission.

### DIFF
--- a/src/main/java/org/folio/rest/impl/TenantPermsAPI.java
+++ b/src/main/java/org/folio/rest/impl/TenantPermsAPI.java
@@ -94,44 +94,42 @@ perm.setModuleName(moduleId.getProduct());
       Context vertxContext, String tenantId) {
     return getPermsByModule(moduleId, vertxContext, tenantId)
         .compose(existing -> {
-          if (existing.isEmpty() && perms != null) {
-            // this means either:
-            // A) the first time enabling this module, or
-            // B) the permissions exist but don't yet have the moduleName field
-            List<Permission> ret = new ArrayList<>();
-            List<Future> futures = new ArrayList<>(perms.size());
-            perms.forEach(perm ->
-                futures.add(getModulePermByName(perm.getPermissionName(), null, vertxContext, tenantId)
-                    .compose(dbPerm -> {
-                      if (dbPerm == null || Boolean.TRUE.equals(dbPerm.getDummy())) {
-                        // permission does not already exist or is dummy
-                        return Future.succeededFuture();
-                      }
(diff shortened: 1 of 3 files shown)


src/main/java/org/folio/rest/impl/TenantPermsAPI.java
```diff
@@ -94,21 +94,28 @@ public void postTenantpermissions(OkapiPermissionSet entity, Map<String, String>
...
-            perms.forEach(perm ->
+          if (!existing.isEmpty()) {
+            return Future.succeededFuture(existing); // permission already has context
+          }
+          // this means either:
+          // A) the first time enabling this module, or
+          // B) the permissions exist but don't yet have the moduleName field
+          List<Permission> ret = new ArrayList<>();
+          List<Future> futures = new ArrayList<>(perms.size());
+          perms.forEach(perm ->
```
craigmcnally, line 105: possible NPE if perms == null...  IIRC there were unit tests which exercised the ability for perms == null, though in practice I don't think it's likely to happen.

src/test/java/org/folio/permstest/RestVerticleTest.java
```diff
@@ -328,6 +328,123 @@ public void testTenantPermissionsNullPermList(TestContext context) {
     context.assertEquals(201, response.code);
   }
 
+  // Test that a permission can be used in another module if first one is deleted MODPERMS-130
```
craigmcnally, line 331: Maybe add to this comment that one situation where this might occur is when two modules implement the same interface.

src/main/java/org/folio/rest/impl/TenantPermsAPI.java
adamdickmeiss: It's not going to happen as `handlePermLists` passes a good list. All parameters to `getPermsForModule` are expected to be non-null, but don't check for those either (and shouldn't).

src/test/java/org/folio/permstest/RestVerticleTest.java
adamdickmeiss: Sure. I've added something.

hjiebsco (approved):
As discussed, re-use perm will benefit the case where different modules implement the same interface with the same permission, which is much more likely than the edge case where the same perm is defined by unrelated modules. We will take the very small risk and get this approved and merged.

craigmcnally (approved):

PR #113, opened 2021-02-10
MODPERMS-129 Deprecate system-defined permissions in tenant init

craigmcnally:
## Purpose
In order to solve the problem of "orphaned" permissions, when mod-permissions is enabled/upgraded, mark all system-defined permissions (those w/ mutable=false) as deprecated.  As OKAPI calls _/tenantPermissions for all the enabled modules, the permissions which are still in use will be re-enabled (marked deprecated=false).  Any permissions remaining would have been orphaned, but are now deprecated, and can be cleaned up.

See https://issues.folio.org/browse/MODPERMS-129

## Approach
* Implemented this in TenantRefApi.loadData(...) since schemas/tables/etc. should all be setup at that point.  
  * Retrieve all permissions w/ mutable=false
  * For each, set deprecated=true and prepend `(deprecated)` to the displayName
  * Update the database
* Implement a unit test to exercise this and ensure it doesn't affect user-defined permissions
* Update TenantPermsAPI as needed - there were a few checks which were preventing the deprecated permissions from being picked up and acted on.  In most cases the fact that we changed the displayName to have the deprecated prefix was sufficient, but relying on that seemed inelegant so it now explicitly checks the deprecated field in these places.

## Notes
* I pulled in the changes made for MODPERMS-130 (https://github.com/folio-org/mod-permissions/pull/112) as I think they were needed for this to work.

### DIFF
--- a/src/main/java/org/folio/rest/impl/TenantPermsAPI.java
+++ b/src/main/java/org/folio/rest/impl/TenantPermsAPI.java
@@ -3,6 +3,14 @@ package org.folio.rest.impl;
 import static org.folio.rest.impl.PermsAPI.checkPermissionExists;
 import static org.folio.rest.impl.PermsAPI.getCQL;
 import static org.folio.rest.impl.PermsAPI.getIdCriterion;
+import io.vertx.core.AsyncResult;
+import io.vertx.core.CompositeFuture;
+import io.vertx.core.Context;
+import io.vertx.core.Future;
+import io.vertx.core.Handler;
+import io.vertx.core.Promise;
+import io.vertx.core.json.Json;
+import io.vertx.core.json.JsonArray;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
@@ -10,6 +18,8 @@ import java.util.Map;
 import java.util.UUID;
 import java.util.stream.Collectors;
 import javax.ws.rs.core.Response;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.folio.okapi.common.ModuleId;
 import org.folio.okapi.common.SemVer;
 import org.folio.rest.jaxrs.model.OkapiPermission;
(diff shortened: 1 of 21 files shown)


craigmcnally:
Hmm I'm thinking maybe I should have rebased.

adamdickmeiss:
> Hmm I'm thinking maybe I should have rebased.

MODPERMS-130 is now merged. I don't it's a problem merge-wise with this PR even based on MODPERMS-130.

Will try out this PR tomorrow.

src/main/java/org/folio/rest/impl/TenantPermsAPI.java
```diff
@@ -718,7 +720,8 @@ public void postTenantpermissions(OkapiPermissionSet entity, Map<String, String>
                 return;
               }
```
adamdickmeiss, line 721: I have an issue with the preceding condition that checks for collision. Iv'e seen it reporting collision for mutable=false and also deprecated permission. For this reason (and as you did below),. I'd wisth we changed it to:
```java
         if (foundPerm.getModuleName() == null && !Boolean.TRUE.equals(foundPerm.getDeprecated())) {
             pgClient.rollbackTx(connection, rollback ->
            promise.fail("PermissionName collision with user-defined permission: " + perm.getPermissionName()));
         }
```

With this in mind the logic could be simplified.

adamdickmeiss (commented):
It looks good.. Just one wish for the collision check when saving a permission.

src/main/java/org/folio/rest/impl/TenantPermsAPI.java
craigmcnally: Can we make this change in scope of another bug/PR?

adamdickmeiss: Yes

adamdickmeiss (approved):
Code looks good to me. This will essentially mean that Okapi is required to refresh all modules for this tenant version = 2.0.

PR #114, opened 2021-02-11
Do not assign context to dummy permissions MODPERMS-133

adamdickmeiss:
Dummy permissions should not be assigned a context. They don't belong to the module where they happen to be referred to.

They should be assigned a context when they are really defined.

Furthermore the collision check in savePerms is wrong for. It checks if context is null and then assumes it's user-defined context.
If a permission has Mutable undefined or TRUE – then it'a user-defined permission.

This PR also makes it possible to use stock install.json for Okapi. OKAPI-985.

### DIFF
--- a/src/main/java/org/folio/rest/impl/TenantPermsAPI.java
+++ b/src/main/java/org/folio/rest/impl/TenantPermsAPI.java
@@ -721,7 +721,8 @@ perm.setModuleName(moduleId.getProduct());
               if ((perm.getSubPermissions() != null && !perm.getSubPermissions().equals(foundPerm.getSubPermissions()))
                   || (perm.getVisible() != null && !perm.getVisible().equals(foundPerm.getVisible()))
                   || (perm.getDisplayName() != null && !perm.getDisplayName().equals(foundPerm.getDisplayName()))
-                  || (perm.getDescription() != null && !perm.getDescription().equals(foundPerm.getDescription()))) {
+                  || (perm.getDescription() != null && !perm.getDescription().equals(foundPerm.getDescription()))
+                  || (moduleId.getSemVer() != null && !moduleId.getSemVer().toString().equals(foundPerm.getModuleVersion()))) {
                 foundPerm.setDummy(true);
               }
             }
--- a/src/main/java/org/folio/rest/impl/TenantPermsAPI.java
+++ b/src/main/java/org/folio/rest/impl/TenantPermsAPI.java
@@ -712,7 +712,7 @@ perm.setModuleName(moduleId.getProduct());
             Permission foundPerm = null;
             if (!returnList.isEmpty()) {
               foundPerm = returnList.get(0);
-              if (foundPerm.getModuleName() == null) {
+              if (!Boolean.FALSE.equals(foundPerm.getMutable())) {
                 pgClient.rollbackTx(connection, rollback ->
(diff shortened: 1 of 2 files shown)


craigmcnally:
> Dummy permissions should not be assigned a context. They don't belong to the module where they happen to be referred to.
They should be assigned a context when they are really defined.

I agree - this is a good catch.

craigmcnally:
> Furthermore the collision check in savePerms is wrong for. It checks if context is null and then assumes it's user-defined context.
If a permission has Mutable undefined or TRUE – then it'a user-defined permission.

As we discussed before we didn't want to rely on mutable as the indicator that a permission is system-defined because it's currently possible for a user to create an immutable permission.

That said, your changes in MODPERMS-130 already go down this path, so I guess we've changed our mind here and should just be consistent.

adamdickmeiss:
> > Furthermore the collision check in savePerms is wrong for. It checks if context is null and then assumes it's user-defined context.
> > If a permission has Mutable undefined or TRUE – then it'a user-defined permission.
> 
> As we discussed before we didn't want to rely on mutable as the indicator that a permission is system-defined because it's currently possible for a user to create an immutable permission.
> 
> That said, your changes in MODPERMS-130 already go down this path, so I guess we've changed our mind here and should just be consistent.

Here's a check without mutable .. this does not change test results.
```java
 if (!Boolean.TRUE.equals(foundPerm.getDummy()) && foundPerm.getModuleName() == null) {
```

I'm not sure it's prettier.. We have an existing database with Dummy (not moduleName).

craigmcnally:
When I implemented the static permission migration stuff the agreement was at the time to upgrade mod-permissions first, which in conjunction with the latest OKAPI code would result in all system-defined perms getting module context.  In that case using moduleName as an indicator that a permission was system or user defined was appropriate.  For various reasons, a lot has changed since then, including the agreement to not relying on mutable as that indicator...

adamdickmeiss:
> When I implemented the static permission migration stuff the agreement was at the time to upgrade mod-permissions first, which in conjunction with the latest OKAPI code would result in all system-defined perms getting module context. In that case using moduleName as an indicator that a permission was system or user defined was appropriate. For various reasons, a lot has changed since then, including the agreement to not relying on mutable as that indicator...

It does seem like we're stuck with mutable as an indicator. Just merged PR https://github.com/folio-org/mod-permissions/pull/113  assumes exactly that mutable is the indicator and transforms system perms to Deprecated. And not just for from version 1.1 to 2 - any time mod-permissions is enabled, if I read the PR correctly.

craigmcnally:
> > When I implemented the static permission migration stuff the agreement was at the time to upgrade mod-permissions first, which in conjunction with the latest OKAPI code would result in all system-defined perms getting module context. In that case using moduleName as an indicator that a permission was system or user defined was appropriate. For various reasons, a lot has changed since then, including the agreement to not relying on mutable as that indicator...
> 
> It does seem like we're stuck with mutable as an indicator. Just merged PR #113 assumes exactly that mutable is the indicator and transforms system perms to Deprecated. And not just for from version 1.1 to 2 - any time mod-permissions is enabled, if I read the PR correctly.

Right, though I think we're also planning on making mutable a read-only field so this should be a one-time risk.  Going forward, users shouldn't be able to create immutable permissions.

hjiebsco (approved):

PR #115, opened 2021-02-23
MODPERMS-126 user-defined perms are always mutable

craigmcnally:
## Purpose
The original intent of the mutable field was to serve as an indicator that a permission was defined by the system.  However, for whatever reason, the property was not treated read-only.  This means while it was never the intention to allow users to create immutable permissions (those which cannot be modified/removed), it's technically possible.

This PR removes the mutable field from the permissionUpload.json schema.  

This is a breaking change.

*N.B.* Any user-defined, immutable permissions that already exist will be deprecated upon upgrade to Iris.

See https://issues.folio.org/browse/MODPERMS-126

### DIFF
--- a/src/main/java/org/folio/rest/impl/PermsAPI.java
+++ b/src/main/java/org/folio/rest/impl/PermsAPI.java
@@ -1896,7 +1896,6 @@ public class PermsAPI implements Perms {
       subPerms.add(s);
     }
     perm.setSubPermissions(subPerms);
-    perm.setMutable(entity.getMutable());
     perm.setVisible(entity.getVisible());
     perm.setTags(entity.getTags());
     perm.setMetadata(entity.getMetadata());
--- a/src/main/java/org/folio/rest/impl/PermsAPI.java
+++ b/src/main/java/org/folio/rest/impl/PermsAPI.java
@@ -1897,6 +1897,7 @@ public class PermsAPI implements Perms {
     }
     perm.setSubPermissions(subPerms);
     perm.setVisible(entity.getVisible());
+    perm.setMutable(true); // user-defined perms are always mutable
     perm.setTags(entity.getTags());
     perm.setMetadata(entity.getMetadata());
     return perm;
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -4,7 +4,7 @@
   "provides": [
     {
       "id": "permissions",
-      "version": "5.3",
+      "version": "6.0",
       "handlers" : [
         {
           "methods": [ "POST" ],
--- a/ramls/permissionUpload.json
+++ b/ramls/permissionUpload.json
@@ -34,10 +34,6 @@
         "type": "string"
       }
     },
-    "mutable": {
-      "description": "Whether or not this permission can change at runtime",
-      "type": "boolean"
-    },
     "visible": {
       "description": "Should this permission be hidden from users",
       "type": "boolean"
--- a/src/test/java/org/folio/permstest/RestVerticleTest.java
+++ b/src/test/java/org/folio/permstest/RestVerticleTest.java
@@ -918,7 +918,12 @@ public class RestVerticleTest {
     context.assertEquals(response.code, 201);
     context.assertEquals(uuid, response.body.getString("id")); // MODPERMS-84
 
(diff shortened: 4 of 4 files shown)


src/main/java/org/folio/rest/impl/PermsAPI.java
```diff
@@ -1896,7 +1896,6 @@ private static Permission getRealPermObject(PermissionUpload entity) {
       subPerms.add(s);
     }
     perm.setSubPermissions(subPerms);
-    perm.setMutable(entity.getMutable());
     perm.setVisible(entity.getVisible());
```
hjiebsco, line 1899: When can `mutable` be true then? Maybe `perm.setMutable(true)` here since these are user defined perms?

hjiebsco (commented):
Left an inline comment about when `mutable` can be true. Also, we need bump up major interface version since it is a breaking change.

src/main/java/org/folio/rest/impl/PermsAPI.java
adamdickmeiss: Yes. Must be set to true here and a unit test should test this this indeed the case. Eg you could add to `RestVerticleTest.testPostPermsPermissions`:

```java
   response = send(HttpMethod.GET, "/perms/permissions/" + uuid, null, context);
   context.assertEquals(response.code, 200);
   context.assertTrue(Boolean.TRUE.equals(response.body.getBoolean("mutable")), response.body.encodePrettily());
```

adamdickmeiss (changes requested):

src/main/java/org/folio/rest/impl/PermsAPI.java
craigmcnally: `mutable` isn't a required field, so I figured we'd just omit it, but I'm fine with making it explicit.  I'll fix this in a moment.

craigmcnally: In any case you're right that a unit test should be updated/added.

adamdickmeiss (approved):
Thx

hjiebsco (approved):
I assume we still need to bump up the interface major version.

adamdickmeiss:
> I assume we still need to bump up the interface major version.

If we think *nothing* is passing mutable at this stage, then I'd let the version stay - just not worth the trouble, IMHO. And should clients be passing it, that would fail.

adamdickmeiss:
> > I assume we still need to bump up the interface major version.
> 
> If we think _nothing_ is passing mutable at this stage, then I'd let the version stay - just not worth the trouble, IMHO. And should clients be passing it, that would fail.

I see this https://github.com/folio-org/folio-ansible/blob/e942650f19f6ff35af5dd92556f435c0fe780e21/doc/ansible-variables.md
that means the ansible playbook assuming mutable being present as a property. Even *if* the interface was updated, something "outside" modules are assuming a structure.

craigmcnally:
> I assume we still need to bump up the interface major version.

yup, I missed this before... Will make the change.  Sounds like we'll need to coordinate merging this with the changes in the ansible playbook, and look for other places that might assume this can be provided.

adamdickmeiss:
> > I assume we still need to bump up the interface major version.
> 
> yup, I missed this before... Will make the change. Sounds like we'll need to coordinate merging this with the changes in the ansible playbook, and look for other places that might assume this can be provided.

Permissions are so fundamental.. One could also just leave the flag as is and force the value upon adding it (simply ignore it).

adamdickmeiss:
> > > I assume we still need to bump up the interface major version.
> > 
> > 
> > yup, I missed this before... Will make the change. Sounds like we'll need to coordinate merging this with the changes in the ansible playbook, and look for other places that might assume this can be provided.
> 
> Permissions are so fundamental.. One could also just leave the flag as is and force the value upon adding it (simply ignore it). We are so close to the deadline.. Breaking this now - will make all other development be stalled - until resolved. Just saying.

Inspecting the folio-ansible more, it appears that this is just a markdown document. It is also mentioned in a deprecated role. So perhaps this is not such a big deal anyway.

craigmcnally:
For reference:

julianladisch (approved):

craigmcnally:
## UPDATE
We discussed as a team and decided to leave the schema unchanged, but change the code to ignore `mutable` in the perms API.  Permissions created via this API will always be mutable.  I'm going to close this PR and open a new one with those changes.

PR #116, opened 2021-02-24
MODPERMS-116 purge deprecated perms

hjiebsco:
Please see https://issues.folio.org/browse/MODPERMS-116 Create purge Deprecated API

Some minor changes from original JIRA description

- endpoint was changed from `/perms/permissions/purge-deprecated` to `/perms/purge-deprecated` because of conflict with `/perms/permissions/{id}`
- reused the existing [permissionNameListObject.json](https://github.com/folio-org/mod-permissions/blob/master/ramls/permissionNameListObject.json) rather than creating a new json to define the response. They have the exact same structure except property naming

### DIFF
--- a/ramls/permissions.raml
+++ b/ramls/permissions.raml
@@ -415,4 +415,17 @@ traits:
             body:
               text/plain:
                 example: "Internal server error"
-
+  /purge-deprecated:
+    post:
+      description: purge deprecated permissions
+      responses:
+        200:
+          description: "A list of permission names purged"
+          body:
+            application/json:
+              type: permissionNameListObject
+        500:
+          description: "Internal server error"
+          body:
+            text/plain:
+              example: "Internal server error"
--- a/ramls/permissions.raml
+++ b/ramls/permissions.raml
@@ -424,6 +424,11 @@ traits:
           body:
             application/json:
               type: permissionNameListObject
+        400:
+          description: "Bad request"
+          body:
+            text/plain:
+              example: "Bad request"
         500:
           description: "Internal server error"
           body:
--- a/src/main/java/org/folio/rest/impl/PermissionUtils.java
+++ b/src/main/java/org/folio/rest/impl/PermissionUtils.java
@@ -9,7 +9,6 @@ import org.folio.rest.jaxrs.model.OkapiPermission;
 import org.folio.rest.jaxrs.model.Permission;
 import org.folio.rest.jaxrs.model.PermissionNameListObject;
 import org.folio.rest.persist.PostgresClient;
-import io.vertx.core.CompositeFuture;
 import io.vertx.core.Future;
 import io.vertx.core.Promise;
 import io.vertx.sqlclient.Row;
--- a/src/main/java/org/folio/rest/impl/PermissionUtils.java
+++ b/src/main/java/org/folio/rest/impl/PermissionUtils.java
@@ -76,6 +76,10 @@ public class PermissionUtils {
     pgClient.startTx(tx -> {
       try {
         pgClient.select(tx, String.format(SELECT_DEPRECATED_PERMS, tenantId), res -> {
+          if (res.failed()) {
+            pgClient.rollbackTx(tx, done -> promise.fail(res.cause()));
+            return;
+          }
           PermissionNameListObject permNames = new PermissionNameListObject();
           permNames.setTotalRecords(0);
           @SuppressWarnings("rawtypes")
@@ -92,17 +96,14 @@ public class PermissionUtils {
             });
           CompositeFuture.all(futures)
             .onSuccess(ignore -> {
-              pgClient.endTx(tx, done -> done.mapEmpty());
-              promise.complete(permNames);
+              pgClient.endTx(tx, done -> promise.complete(permNames));
             })
             .onFailure(ex -> {
-              pgClient.rollbackTx(tx, done -> done.mapEmpty());
-              promise.fail(ex);
+              pgClient.rollbackTx(tx, done -> promise.fail(ex));
             });
         });
       } catch (Exception e) {
-        pgClient.rollbackTx(tx, done -> done.mapEmpty());
-        promise.fail(e);
+        pgClient.rollbackTx(tx, done -> promise.fail(e));
       }
     });
     return promise.future();
--- a/src/main/java/org/folio/rest/impl/PermissionUtils.java
+++ b/src/main/java/org/folio/rest/impl/PermissionUtils.java
@@ -1,10 +1,30 @@
 package org.folio.rest.impl;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Objects;
 import org.folio.rest.jaxrs.model.OkapiPermission;
 import org.folio.rest.jaxrs.model.Permission;
+import org.folio.rest.jaxrs.model.PermissionNameListObject;
+import org.folio.rest.persist.PostgresClient;
+import io.vertx.core.CompositeFuture;
+import io.vertx.core.Future;
+import io.vertx.core.Promise;
+import io.vertx.sqlclient.Row;
+import io.vertx.sqlclient.RowSet;
 
 public class PermissionUtils {
+
+  private static final String SELECT_DEPRECATED_PERMS = "select id::text as id, jsonb->>'permissionName' as name "
+      + "from %s_mod_permissions.permissions where jsonb->>'deprecated' = 'true'";
+  private static final String PURGE_DEPRECATED_PERMS = "delete from %s_mod_permissions.permissions "
+      + "where jsonb->>'deprecated' = 'true'";
+  private static final String PURGE_DEPRECATED_SUB_PERMS = "update %s_mod_permissions.permissions "
+      + "set jsonb = jsonb_set(jsonb, '{subPermissions}', (jsonb->'subPermissions')::jsonb - array['%s', '%s']) "
+      + "where jsonb->'subPermissions' ?| array['%s', '%s']";
+  private static final String PURGE_DEPRECATED_PERMS_USERS = "update %s_mod_permissions.permissions_users "
+      + "set jsonb = jsonb_set(jsonb, '{permissions}', (jsonb->'permissions')::jsonb - array['%s', '%s']) "
+      + "where jsonb->'permissions' ?| array['%s', '%s']";
 
   private PermissionUtils() {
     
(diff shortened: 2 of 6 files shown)


ramls/permissions.raml
```diff
@@ -415,4 +415,17 @@ traits:
...
+        200:
+          description: "A list of permission names purged"
+          body:
+            application/json:
+              type: permissionNameListObject
+        500:
+          description: "Internal server error"
+          body:
+            text/plain:
+              example: "Internal server error"
```
adamdickmeiss, line 431: I would add 400 as a possible status here, so that you at least declare this as a possible outcome. Never know what RMB could do before your code is executed.

src/main/java/org/folio/rest/impl/PermissionUtils.java
```diff
@@ -1,11 +1,31 @@
...
+  private static final String PURGE_DEPRECATED_PERMS = "delete from %s_mod_permissions.permissions "
+      + "where jsonb->>'deprecated' = 'true'";
+  private static final String PURGE_DEPRECATED_SUB_PERMS = "update %s_mod_permissions.permissions "
+      + "set jsonb = jsonb_set(jsonb, '{subPermissions}', (jsonb->'subPermissions')::jsonb - array['%s', '%s']) "
+      + "where jsonb->'subPermissions' ?| array['%s', '%s']";
+  private static final String PURGE_DEPRECATED_PERMS_USERS = "update %s_mod_permissions.permissions_users "
+      + "set jsonb = jsonb_set(jsonb, '{permissions}', (jsonb->'permissions')::jsonb - array['%s', '%s']) "
+      + "where jsonb->'permissions' ?| array['%s', '%s']";
+
   private PermissionUtils() {
```
adamdickmeiss, line 29: LGTM, but not an expert here.

src/main/java/org/folio/rest/impl/PermissionUtils.java
```diff
@@ -42,5 +62,49 @@ public static boolean equals(OkapiPermission okapiPerm, Permission perm) {
...
+   * @param pgClient {@link PostgresClient}
+   * @param tenantId tenant id
+   * @return {@link Future} with {@link PermissionNameListObject}
+   */
+  public static Future<PermissionNameListObject> purgeDeprecatedPermissions(PostgresClient pgClient, String tenantId) {
+    Promise<PermissionNameListObject> promise = Promise.promise();
+    pgClient.startTx(tx -> {
+      try {
+        pgClient.select(tx, String.format(SELECT_DEPRECATED_PERMS, tenantId), res -> {
+          PermissionNameListObject permNames = new PermissionNameListObject();
```
adamdickmeiss, line 79: No check for success/failure here. A bad tenant (no init = no tables) will make a failures and the res.result() below will be an unhandled NPE resulting in no HTTP response generated.
```suggestion
          if (res.failed()) {
            pgClient.rollbackTx(tx, done -> promise.fail(res.cause()));
            return;
          }
          PermissionNameListObject permNames = new PermissionNameListObject();
```

src/main/java/org/folio/rest/impl/PermissionUtils.java
```diff
@@ -42,5 +62,49 @@ public static boolean equals(OkapiPermission okapiPerm, Permission perm) {
...
+              futures.add(Future.<RowSet<Row>>future(p -> pgClient.execute(tx, String.format(PURGE_DEPRECATED_SUB_PERMS, tenantId, id, name, id, name), p)));
+              futures.add(Future.<RowSet<Row>>future(p -> pgClient.execute(tx, String.format(PURGE_DEPRECATED_PERMS_USERS, tenantId, id, name, id, name), p)));
+              futures.add(Future.<RowSet<Row>>future(p -> pgClient.execute(tx, String.format(PURGE_DEPRECATED_PERMS, tenantId), p)));
+              permNames.getPermissionNames().add(name);
+              permNames.setTotalRecords(permNames.getTotalRecords() + 1);
+            });
+          CompositeFuture.all(futures)
+            .onSuccess(ignore -> {
+              pgClient.endTx(tx, done -> done.mapEmpty());
+              promise.complete(permNames);
```
adamdickmeiss, line 96: the done.map is redundant. It is desired that completion is not signalled before endTx/rollback is complete.
```suggestion
              pgClient.endTx(tx, done -> promise.complete(permNames));
```

src/main/java/org/folio/rest/impl/PermissionUtils.java
```diff
@@ -42,5 +62,49 @@ public static boolean equals(OkapiPermission okapiPerm, Permission perm) {
...
+            })
+            .onFailure(ex -> {
+              pgClient.rollbackTx(tx, done -> done.mapEmpty());
+              promise.fail(ex);
```
adamdickmeiss, line 100:
```suggestion
              pgClient.rollbackTx(tx, done -> promise.fail(ex));
```

src/main/java/org/folio/rest/impl/PermissionUtils.java
```diff
@@ -42,5 +62,49 @@ public static boolean equals(OkapiPermission okapiPerm, Permission perm) {
...
+            });
+        });
+      } catch (Exception e) {
+        pgClient.rollbackTx(tx, done -> done.mapEmpty());
+        promise.fail(e);
```
adamdickmeiss, line 105:
```suggestion
        pgClient.rollbackTx(tx, done -> promise.fail(e));
```

src/main/java/org/folio/rest/impl/PermsAPI.java
```diff
@@ -1937,4 +1937,28 @@ private static Criterion buildPermissionNameListQuery(List<String> permissionNam
...
+      PostgresClient pgClient = PostgresClient.getInstance(vertxContext.owner(), tenantId);
+      PermissionUtils.purgeDeprecatedPermissions(pgClient, tenantId)
+        .onSuccess(res -> {
+          asyncResultHandler
+            .handle(Future.succeededFuture(PostPermsPurgeDeprecatedResponse.respond200WithApplicationJson(res)));
+        })
+        .onFailure(ex -> {
+          asyncResultHandler
+            .handle(Future.succeededFuture(PostPermsPurgeDeprecatedResponse.respond500WithTextPlain(ex.getCause()
+              .getMessage())));
```
adamdickmeiss, line 1955: `ex.getCause()` may be null. Use `ex.getMessage()`
```suggestion
          asyncResultHandler
            .handle(Future.succeededFuture(PostPermsPurgeDeprecatedResponse.respond500WithTextPlain(ex.getMessage())));
```

src/test/java/org/folio/rest/impl/PermsAPITest.java
```diff
@@ -207,4 +207,13 @@ public void testGetPermsPermissionsNullPointer(TestContext context) {
...
 
+  @Test
+  public void testPostPermsPurgeDeprecatedNullPointer(TestContext context) {
+    PermsAPI api = new PermsAPI();
+
+    api.postPermsPurgeDeprecated(null, context.asyncAssertSuccess(res -> {
+      context.assertEquals(500, res.getStatus());
+    }), null);
+  }
+
```
adamdickmeiss, line 218:
```suggestion

  @Test
  public void testPostPermsPurgeDeprecatedBadTenant(TestContext context) {
      PermsAPI api = new PermsAPI();
      Map<String,String> headers = new HashMap<>();
      headers.put("x-okapi-tenant", "foo");
      api.postPermsPurgeDeprecated(headers, context.asyncAssertSuccess(res -> {
        context.assertEquals(500, res.getStatus());
      }), vertx.getOrCreateContext());
  }
```

adamdickmeiss (changes requested):
One major problem and a few minor ones.

ramls/permissions.raml
hjiebsco: Make sense. Added.

src/main/java/org/folio/rest/impl/PermissionUtils.java
hjiebsco: I am no expert either. I just did what the pg doc says. :)

src/main/java/org/folio/rest/impl/PermissionUtils.java
hjiebsco: Added. Thanks for the test case.

src/main/java/org/folio/rest/impl/PermissionUtils.java
hjiebsco: Make sense. Changed.

src/main/java/org/folio/rest/impl/PermsAPI.java
hjiebsco: Make sense. Changed.

src/main/java/org/folio/rest/impl/PermissionUtils.java
```diff
@@ -42,5 +62,50 @@ public static boolean equals(OkapiPermission okapiPerm, Permission perm) {
...
+            .forEach(row -> {
+              String id = row.getString("id");
+              String name = row.getString("name");
+              futures.add(Future.<RowSet<Row>>future(p -> pgClient.execute(tx, String.format(PURGE_DEPRECATED_SUB_PERMS, tenantId, id, name, id, name), p)));
+              futures.add(Future.<RowSet<Row>>future(p -> pgClient.execute(tx, String.format(PURGE_DEPRECATED_PERMS_USERS, tenantId, id, name, id, name), p)));
+              futures.add(Future.<RowSet<Row>>future(p -> pgClient.execute(tx, String.format(PURGE_DEPRECATED_PERMS, tenantId), p)));
+              permNames.getPermissionNames().add(name);
+              permNames.setTotalRecords(permNames.getTotalRecords() + 1);
+            });
+          CompositeFuture.all(futures)
```
craigmcnally, line 97: Can we use https://github.com/folio-org/okapi/blob/master/okapi-common/src/main/java/org/folio/okapi/common/GenericCompositeFuture.java here yet?  I think the issues with dependency versions have been resolved w/ the upgrade to RMB 32

src/main/java/org/folio/rest/impl/PermissionUtils.java
```diff
@@ -42,5 +62,50 @@ public static boolean equals(OkapiPermission okapiPerm, Permission perm) {
...
+      try {
+        pgClient.select(tx, String.format(SELECT_DEPRECATED_PERMS, tenantId), res -> {
+          if (res.failed()) {
+            pgClient.rollbackTx(tx, done -> promise.fail(res.cause()));
+            return;
+          }
+          PermissionNameListObject permNames = new PermissionNameListObject();
+          permNames.setTotalRecords(0);
+          @SuppressWarnings("rawtypes")
+          List<Future> futures = new ArrayList<Future>();
```
craigmcnally, line 86: Can we use https://github.com/folio-org/okapi/blob/master/okapi-common/src/main/java/org/folio/okapi/common/GenericCompositeFuture.java here yet? I think the issues with dependency versions have been resolved w/ the upgrade to RMB 32

adamdickmeiss (approved):
As @craigmcnally mentioned, GenericCompositeFuture could be used. I think we should make a separate PR for that to have them "all at once". Up to you.

src/main/java/org/folio/rest/impl/PermissionUtils.java
hjiebsco: Changed to use `GenericCompositeFuture`

adamdickmeiss (approved):

PR #117, opened 2021-02-25
MODPERMS-135 update subperms for perm replace

hjiebsco:
Please see https://issues.folio.org/browse/MODPERMS-135: "replaces" processing doesn't update permissions, only permission assignments

Note:

- Used direct sql in case that subperms do not have matching `childOf` defined
- Also updated earlier commit to use only perm name, not perm id (not needed) to purge deprecated ones

### DIFF
--- a/src/main/java/org/folio/rest/impl/PermissionUtils.java
+++ b/src/main/java/org/folio/rest/impl/PermissionUtils.java
@@ -13,6 +13,7 @@ import io.vertx.core.Future;
 import io.vertx.core.Promise;
 import io.vertx.sqlclient.Row;
 import io.vertx.sqlclient.RowSet;
+import io.vertx.sqlclient.Tuple;
 
 public class PermissionUtils {
 
@@ -21,11 +22,11 @@ public class PermissionUtils {
   private static final String PURGE_DEPRECATED_PERMS = "delete from %s_mod_permissions.permissions "
       + "where jsonb->>'deprecated' = 'true'";
   private static final String PURGE_DEPRECATED_SUB_PERMS = "update %s_mod_permissions.permissions "
-      + "set jsonb = jsonb_set(jsonb, '{subPermissions}', (jsonb->'subPermissions')::jsonb - '%s') "
-      + "where jsonb->'subPermissions' ? '%s'";
+      + "set jsonb = jsonb_set(jsonb, '{subPermissions}', (jsonb->'subPermissions')::jsonb - $1) "
+      + "where jsonb->'subPermissions' ? $2";
   private static final String PURGE_DEPRECATED_PERMS_USERS = "update %s_mod_permissions.permissions_users "
-      + "set jsonb = jsonb_set(jsonb, '{permissions}', (jsonb->'permissions')::jsonb - '%s') "
-      + "where jsonb->'permissions' ? '%s'";
+      + "set jsonb = jsonb_set(jsonb, '{permissions}', (jsonb->'permissions')::jsonb - $1) "
+      + "where jsonb->'permissions' ? $2";
 
   private PermissionUtils() {
     
@@ -87,8 +88,8 @@ public class PermissionUtils {
           res.result()
             .forEach(row -> {
               String name = row.getString("name");
-              futures.add(Future.<RowSet<Row>>future(p -> pgClient.execute(tx, String.format(PURGE_DEPRECATED_SUB_PERMS, tenantId, name, name), p)));
-              futures.add(Future.<RowSet<Row>>future(p -> pgClient.execute(tx, String.format(PURGE_DEPRECATED_PERMS_USERS, tenantId, name, name), p)));
+              futures.add(Future.<RowSet<Row>>future(p -> pgClient.execute(tx, String.format(PURGE_DEPRECATED_SUB_PERMS, tenantId), Tuple.of(name, name), p)));
+              futures.add(Future.<RowSet<Row>>future(p -> pgClient.execute(tx, String.format(PURGE_DEPRECATED_PERMS_USERS, tenantId), Tuple.of(name, name), p)));
               futures.add(Future.<RowSet<Row>>future(p -> pgClient.execute(tx, String.format(PURGE_DEPRECATED_PERMS, tenantId), p)));
               permNames.getPermissionNames().add(name);
               permNames.setTotalRecords(permNames.getTotalRecords() + 1);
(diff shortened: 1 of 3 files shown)


craigmcnally (approved):
LGTM

src/test/java/org/folio/permstest/RestVerticleTest.java
```diff
@@ -617,9 +620,68 @@ public void testPurgeDeprecatedPerm(TestContext context) {
...
+    // verify perm3 has the perm4 in subperm
+    response = send(HttpMethod.GET, "/perms/permissions?query=permissionName%3D" + perm3, null, context);
+    context.assertEquals(200, response.code);
+    JsonObject permObject = response.body.getJsonArray("permissions").getJsonObject(0);
+    context.assertTrue(permObject.getJsonArray("subPermissions").contains(perm4));
+
+    // verify permuser has perm4 assigned
+    response = send(HttpMethod.GET, "/perms/users/" + permUserId, null, context);
+    JsonObject permUsersObject = response.body;
+    context.assertTrue(permUsersObject.getJsonArray("permissions").contains(perm4));
```
julianladisch, line 684: context.assert*(...) can be replaced by assertThat(...) if
* it is not in a callback (here), or
* it is in a callback and wrapped in context.asyncAssertSuccess, context.asyncAssertFailure or context.vertify.
assertThat increases readability for both code and error messages.

julianladisch, line 684: Can you check the complete JsonArray?

```

    JsonArray abd = new JsonArray().add(perm1).add(perm2).add(perm4);

    // verify perm3 has the perm4 in subperm
    response = send(HttpMethod.GET, "/perms/permissions?query=permissionName%3D" + perm3, null, context);
    assertThat(response.body.getJsonArray("permissions")
        .getJsonObject(0).getJsonArray("subPermissions"), is(abd));

    // verify permuser has perm4 assigned
    response = send(HttpMethod.GET, "/perms/users/" + permUserId, null, context);
    assertThat(response.body.getJsonArray("permissions"), is(abd));
```

src/test/java/org/folio/permstest/RestVerticleTest.java
```diff
@@ -617,9 +620,68 @@ public void testPurgeDeprecatedPerm(TestContext context) {
...
+  }
+
+  @Test
+  public void testPermissionNameReplace(TestContext context) {
+    // clean up deprecated perms first
+    Response response = send(HttpMethod.POST, "/perms/purge-deprecated", null, context);
+
+    // seed modA with perm1 and perm2
+    String perm1 = "permA" + UUID.randomUUID().toString();
+    String perm2 = "permB" + UUID.randomUUID().toString();
```
julianladisch, line 634: Can you rename perm1, perm2, perm3, perm4 to permA, permB, permC, permD?

src/test/java/org/folio/permstest/RestVerticleTest.java
```diff
@@ -617,9 +620,68 @@ public void testPurgeDeprecatedPerm(TestContext context) {
...
+    JsonObject permissionSet = new JsonObject()
+        .put("moduleId","moduleTestReplaceA-1.0.0")
```
julianladisch, line 636: Can we rename the modules to moduleTestReplaceX and moduleTestReplaceY? A and B are already used for permission names.

src/main/java/org/folio/rest/impl/PermissionUtils.java
```diff
@@ -86,10 +86,9 @@ public static boolean equals(OkapiPermission okapiPerm, Permission perm) {
           List<Future<RowSet<Row>>> futures = new ArrayList<Future<RowSet<Row>>>();
           res.result()
             .forEach(row -> {
-              String id = row.getString("id");
               String name = row.getString("name");
-              futures.add(Future.<RowSet<Row>>future(p -> pgClient.execute(tx, String.format(PURGE_DEPRECATED_SUB_PERMS, tenantId, id, name, id, name), p)));
-              futures.add(Future.<RowSet<Row>>future(p -> pgClient.execute(tx, String.format(PURGE_DEPRECATED_PERMS_USERS, tenantId, id, name, id, name), p)));
+              futures.add(Future.<RowSet<Row>>future(p -> pgClient.execute(tx, String.format(PURGE_DEPRECATED_SUB_PERMS, tenantId, name, name), p)));
+              futures.add(Future.<RowSet<Row>>future(p -> pgClient.execute(tx, String.format(PURGE_DEPRECATED_PERMS_USERS, tenantId, name, name), p)));
               futures.add(Future.<RowSet<Row>>future(p -> pgClient.execute(tx, String.format(PURGE_DEPRECATED_PERMS, tenantId), p)));
```
julianladisch, line 92: String.format should be replaced by the PostgresClient.execute method that takes `Tuple params`.

src/main/java/org/folio/rest/impl/TenantPermsAPI.java
```diff
@@ -392,8 +398,15 @@ public void postTenantpermissions(OkapiPermissionSet entity, Map<String, String>
...
         .compose(v -> {
           List<Future> futures = new ArrayList<>(permList.size());
+          PostgresClient pgClient = PostgresClient.getInstance(vertxContext.owner(), tenantId);
           permList.keySet().forEach(okapiPerm -> {
+            String newPermName = okapiPerm.getPermissionName();
             permList.get(okapiPerm).forEach(replaced -> {
+              // add new permission name to all relevant sub permissions
+              String oldPermName = replaced.getPermissionName();
+              futures.add(Future.<RowSet<Row>>future(p -> pgClient.execute(connection,
+                  String.format(ADD_PERM_TO_SUB_PERMS, tenantId, newPermName, oldPermName, newPermName), p)));
```
julianladisch, line 408: String.format should be replaced by the PostgresClient.execute method that takes `Tuple params`.

src/test/java/org/folio/permstest/RestVerticleTest.java
hjiebsco: Changed to use `assertThat` for new code

hjiebsco: Thanks for the example. Done.

src/test/java/org/folio/permstest/RestVerticleTest.java
hjiebsco: Renamed.

src/test/java/org/folio/permstest/RestVerticleTest.java
hjiebsco: Renamed.

src/main/java/org/folio/rest/impl/PermissionUtils.java
hjiebsco: Changed to use `Tuple`. I should have done that initially.

PR #118, opened 2021-02-26
MODPERMS-126 fix handling of mutable

craigmcnally:
## Purpose 
The original intent of the mutable field was to serve as an indicator that a permission was defined by the system. However, for whatever reason, the property was not treated read-only. This means while it was never the intention to allow users to create immutable permissions (those which in theory cannot be modified/removed), it's technically possible.

It was decided to address this by ignoring the `mutable` property in the Perms API.  All user-defined perms created/modified via this API will be mutable.

While implementing this fix it was discovered that permissions marked as immutable can actually still be modified and deleted via the Perms API.  This PR also addresses that glaring issue.

## Approach
* ignore `mutable` in the Perms API (POST/PUT)
* return a 400 when attempting to modify or delete a system defined (immutable) permission.
* update tests to exercise these scenarios
  * Introduce a new test for various aspects of mutable handling
  * remove `RestVerticleTest.testPutPermsPermissionsByIdDummy` since dummy perms will always be immutable.  The new test makes this one redundant, and much of it would need to be rewritten if we were to keep it.
  * Update other tests which attempted to cleanup system-generated perms via `DELETE /perms/permissions`.  Instead use `POST /_/tenantPermissions` w/ empty perms array to cleanup.

## Notes
See https://issues.folio.org/browse/MODPERMS-126
See also #115 which addressed part of this in a different way before we decided to go this route.

### DIFF
--- a/src/test/java/org/folio/permstest/RestVerticleTest.java
+++ b/src/test/java/org/folio/permstest/RestVerticleTest.java
@@ -986,7 +986,7 @@ public class RestVerticleTest {
   @Test
   public void testPostPermsPermissions(TestContext context) {
     String uuid = UUID.randomUUID().toString();
-    String permRequest = "{\"id\": \"" + uuid + "\", \"permissionName\":\"adname\",\"displayName\":\"addisplay\", \"mutable\":false}";
+    String permRequest = "{\"id\": \"" + uuid + "\", \"permissionName\":\"adname\",\"displayName\":\"addisplay\"}";
 
     Response response = send(HttpMethod.POST, "/perms/permissions",
         permRequest, context);
--- a/src/test/java/org/folio/permstest/RestVerticleTest.java
+++ b/src/test/java/org/folio/permstest/RestVerticleTest.java
@@ -450,6 +450,72 @@ public class RestVerticleTest {
     context.assertEquals(400, response.code);
   }
 
+  @Test
+  public void testDummyPerm(TestContext context) {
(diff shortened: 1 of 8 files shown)


adamdickmeiss (approved):

src/test/java/org/folio/permstest/RestVerticleTest.java
```diff
@@ -1043,7 +986,7 @@ public void testDeletePermsUsersPermissionsByIdAndPermissionnameMissingPermissio
   @Test
   public void testPostPermsPermissions(TestContext context) {
     String uuid = UUID.randomUUID().toString();
-    String permRequest = "{\"id\": \"" + uuid + "\", \"permissionName\":\"adname\",\"displayName\":\"addisplay\"}";
+    String permRequest = "{\"id\": \"" + uuid + "\", \"permissionName\":\"adname\",\"displayName\":\"addisplay\", \"mutable\":false}";
```
hjiebsco, line 989: Do not need this change. You new test below covers `PUT` with `mutable: false` case.

hjiebsco (approved):

src/test/java/org/folio/permstest/RestVerticleTest.java
craigmcnally: yup, reverted.  I think this was an artifact of me changing my mind part way through implementing these tests.

PR #127, opened 2021-05-11
MODPERMS-148: RMB 33.0.0-pre3 and Vert.x 4.0.3

julianladisch:


adamdickmeiss (commented):
Why don't we just wait for pre3 whic is in PR.

src/test/java/org/folio/permstest/TestUtil.java
```diff
@@ -123,4 +124,21 @@ public JsonObject getJson() {
...
+    return tenantClient.postTenant(ta)
+    .compose(res -> {
+      if (res.statusCode() == 204) {
+        return Future.succeededFuture();
+      }
+      return Future.failedFuture(new AssertionError("purge: expecting status code 204 but got " + res.statusCode()));
+    });
+  }
+
 }
```
adamdickmeiss, line 144: Could be using TenantInit.purge from RMB here. TenantInit could probably also be used in other places.

julianladisch: Yes.

adamdickmeiss (approved):

PR #128, opened 2021-05-14
Fix leaks SQL connections MODPERMS-150

adamdickmeiss:
Use 33.0.0-pre1 and Vert.x 4.0.2.


adamdickmeiss:
Will use pre4 instead.

PR #143, opened 2021-11-12
Fix perms...immutable can give okapi.all MODPERMS-161

adamdickmeiss:
Require perms.users.assign.okapi to be able to add permissions
any perms.users.assign.okapi + okapi.* for a user. Thus,
perms.users.assign.immutable does not suffice anymore.


julianladisch (approved):
Looks good, thanks!

PR #151, opened 2021-12-30
Improve diagnostics for operating check MODPERMS-175

adamdickmeiss:
Collect all permissions that are failing (rather than report first failure)
Report the modulePermissions as seen from mod-permissions.

### DIFF
--- a/src/main/java/org/folio/rest/impl/PermsAPI.java
+++ b/src/main/java/org/folio/rest/impl/PermsAPI.java
@@ -1024,8 +1024,8 @@ public class PermsAPI implements Perms {
       List<String> modulePermissions) {
     if (!failedPerms.isEmpty()) {
       throw new OperatingUserException(
-          "Cannot add " + type + " permission" + (failedPerms.size() > 1 ? "s" : "")
-              + " " + String.join(", ", failedPerms) + " not owned by operating user "
+          "Cannot add " + type + (failedPerms.size() > 1 ? " permissions " : " permission ")
+              + String.join(", ", failedPerms) + " not owned by operating user "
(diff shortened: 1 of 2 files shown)


MikhailFokanov (approved):
Looks good.

src/main/java/org/folio/rest/impl/PermsAPI.java
```diff
@@ -1020,6 +1020,17 @@ static JsonObject getPayloadWithoutValidation(String token) {
     }
   }
 
+  static void checkPermList(String type, List<String> failedPerms, String operatingUser,
+      List<String> modulePermissions) {
+    if (!failedPerms.isEmpty()) {
+      throw new OperatingUserException(
+          "Cannot add " + type + " permission" + (failedPerms.size() > 1 ? "s" : "")
+              + " " + String.join(", ", failedPerms) + " not owned by operating user "
```
julianladisch, line 1028:
```suggestion
          "Cannot add " + type + (failedPerms.size() > 1 ? " permissions " : " permission ")
              + String.join(", ", failedPerms) + " not owned by operating user "
```

src/main/java/org/folio/rest/impl/PermsAPI.java
```diff
@@ -1020,6 +1020,17 @@ static JsonObject getPayloadWithoutValidation(String token) {
...
+              + operatingUser + (modulePermissions == null || modulePermissions.isEmpty() ?
+              "" : " modulePermissions: " + String.join(", ", modulePermissions)));
```
julianladisch, line 1030:
```suggestion
              "" : ", modulePermissions: " + String.join(", ", modulePermissions)));
```

julianladisch (approved):
Looks good, thanks!

PR #152, opened 2022-01-03
Check for token to check for auth enabled MODPERMS-174

adamdickmeiss:
Previously, if X-Okapi-User-Id was present mod-permissions would
use that as an indication that auth was enabled for the tenant. This
is indeed the case if operating user is the same as tenant user.
Fix by check for X-Okapi-Token being present.


julianladisch (approved):
Looks good, thanks!

PR #159, opened 2022-03-09
Delete bad users, add constraint MODPERMS-177

adamdickmeiss:
Try to keep the table name definitions in one place at least in Java.

The code that saves bad users is not yet discovered. This is just a repair, that will make it harder, but not impossible
to create bad users again (for example a JSONB with userId and additional property such as tags).

### DIFF
--- a/src/test/java/org/folio/permstest/RestVerticleTest.java
+++ b/src/test/java/org/folio/permstest/RestVerticleTest.java
@@ -30,11 +30,13 @@ import org.folio.postgres.testing.PostgresTesterContainer;
 import org.folio.rest.RestVerticle;
 import org.folio.rest.client.TenantClient;
 import org.folio.rest.impl.PermissionUtils;
+import org.folio.rest.impl.PermsAPI;
 import org.folio.rest.impl.PermsCache;
 import org.folio.rest.impl.TenantPermsAPI;
 import org.folio.rest.jaxrs.model.OkapiPermission;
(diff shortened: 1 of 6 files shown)


adamdickmeiss:
The SQ checks is on old code as far I can tell. Should be dealt with separately, I think.

src/test/java/org/folio/permstest/RestVerticleTest.java
```diff
@@ -3119,4 +3121,11 @@ public void permsUsersOffsetLimit(TestContext context) {
     context.assertEquals(200, response.code);
     context.assertEquals(3, response.body.getJsonArray("permissionUsers").size());
   }
+
+  @Test
+  public void testUserIdNull(TestContext context) {
+    PermissionUser u = new PermissionUser().withId(UUID.randomUUID().toString());
+    PostgresClient.getInstance(vertx, "diku").save(PermsAPI.TABLE_NAME_PERMSUSERS, u)
+        .onComplete(context.asyncAssertFailure(e -> context.assertTrue(e.getMessage().contains("permissions_users_userid_not_null"))));
```
julianladisch, line 3129:
```suggestion
        .onComplete(context.asyncAssertFailure(e -> assertThat(e.getMessage(), containsString("permissions_users_userid_not_null"))));
```

julianladisch (approved):
Looks good, thanks!

hjiebsco (approved):
LGTM

PR #162, opened 2022-03-14
Migrate perms.users.item.{post,put} MODPERMS-182

adamdickmeiss:
Also expand permissions to search for these in migration.
Make an adjustment to the Permission User's permissions
property to reflect reality: array of string.

### DIFF
--- a/src/main/java/org/folio/rest/impl/PermissionUtils.java
+++ b/src/main/java/org/folio/rest/impl/PermissionUtils.java
@@ -33,7 +33,7 @@ public class PermissionUtils {
   public static final String PERMS_USERS_ASSIGN_IMMUTABLE = "perms.users.assign.immutable";
   public static final String PERMS_USERS_ASSIGN_MUTABLE = "perms.users.assign.mutable";
   public static final String PERMS_USERS_ASSIGN_OKAPI = "perms.users.assign.okapi";
-  public static final String PERMS_PERMS_ALL = "perms.all";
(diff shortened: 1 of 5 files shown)


src/main/java/org/folio/rest/impl/PermissionUtils.java
```diff
@@ -33,8 +33,8 @@
   public static final String PERMS_USERS_ASSIGN_IMMUTABLE = "perms.users.assign.immutable";
   public static final String PERMS_USERS_ASSIGN_MUTABLE = "perms.users.assign.mutable";
   public static final String PERMS_USERS_ASSIGN_OKAPI = "perms.users.assign.okapi";
-  public static final String PERMS_PERMS_ALL = "perms.all";
   public static final String PERMS_OKAPI_ALL = "okapi.all";
+  protected  static final String [] PERMS_MIGRATE = { "perms.users.item.post", "perms.users.item.put" };
```
julianladisch, line 37:
```suggestion
  protected static final String [] PERMS_MIGRATE = { "perms.users.item.post", "perms.users.item.put" };
```

julianladisch (approved):
Looks good, thanks!

skoczko:
@adamdickmeiss let's hold off on releasing this until we get a chance to discuss. I am no longer convinced this should be merged.

adamdickmeiss:
It is merged, but we'll revert :-)

PR #163, opened 2022-03-14
Release 6.0.2 MODPERMS-183

adamdickmeiss:


skoczko (commented):
@adamdickmeiss Let's hold off on that until we get an agreement on permission sets in Users.

PR #164, opened 2022-03-16
Test user starts including changed permission set

adamdickmeiss:
Just a test that mod-permissions doing the right thing in https://issues.folio.org/browse/UIU-2549


julianladisch (approved):
Looks good, thanks!

Next time consider using RestAssured for readability.

adamdickmeiss:
> Looks good, thanks!
> 
> Next time consider using RestAssured for readability.

Yeah.. mod-permissions uses a utility of its own.. We should switch everywhere.

PR #165, opened 2022-03-22
Fix stack issue with user migration MODPERMS-185

adamdickmeiss:
Fixed by using GenericCompositeFuture rather than compose sequence.

### DIFF
--- a/src/main/java/org/folio/rest/impl/TenantPermsAPI.java
+++ b/src/main/java/org/folio/rest/impl/TenantPermsAPI.java
@@ -261,11 +261,11 @@ public class TenantPermsAPI implements Tenantpermissions {
     CQLWrapper all = new CQLWrapper();
     return connection.get(PermsAPI.TABLE_NAME_PERMSUSERS, PermissionUser.class, all, false)
         .compose(results -> {
-          Future<Void> future = Future.succeededFuture();
+          List<Future<Void>> futures = new ArrayList<>(results.getResults().size());
           for (PermissionUser permUser : results.getResults()) {
-            future = future.compose(x -> migratePermsAssignUser(permUser, connection, vertxContext, tenantId));
+            futures.add(migratePermsAssignUser(permUser, connection, vertxContext, tenantId));
           }
-          return future;
+          return GenericCompositeFuture.all(futures).mapEmpty();
         });
   }
 
--- a/src/test/java/org/folio/permstest/RestVerticleTest.java
+++ b/src/test/java/org/folio/permstest/RestVerticleTest.java
@@ -87,7 +87,7 @@ public class RestVerticleTest {
   static int port;
 
   @Rule
-  public Timeout rule = Timeout.seconds(10);
+  public Timeout rule = Timeout.seconds(10); // set to 900 to test MODPERMS-185
 
   @BeforeClass
   public static void setup(TestContext context) {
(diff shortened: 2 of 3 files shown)


src/test/java/org/folio/permstest/RestVerticleTest.java
```diff
@@ -87,7 +87,7 @@
   static int port;
 
   @Rule
-  public Timeout rule = Timeout.seconds(10);
+  public Timeout rule = Timeout.seconds(10); // set to 900 to test MODPERMS-185
```
julianladisch, line 90:
```
  @Test
  public void testPermsAssignMigration50(TestContext context) {
    testPermsAssignMigration(context, 50);
  }

  @Ignore("Takes many minutes")
  @Test(timeout = 900000l)
  public void testPermsAssignMigration5000(TestContext context) {
    testPermsAssignMigration(context, 5000);
  }

```

src/main/java/org/folio/rest/impl/TenantPermsAPI.java
```diff
@@ -261,11 +261,11 @@ public void postTenantpermissions(OkapiPermissionSet entity, Map<String, String>
...
     return connection.get(PermsAPI.TABLE_NAME_PERMSUSERS, PermissionUser.class, all, false)
         .compose(results -> {
-          Future<Void> future = Future.succeededFuture();
+          List<Future<Void>> futures = new ArrayList<>(results.getResults().size());
           for (PermissionUser permUser : results.getResults()) {
-            future = future.compose(x -> migratePermsAssignUser(permUser, connection, vertxContext, tenantId));
+            futures.add(migratePermsAssignUser(permUser, connection, vertxContext, tenantId));
           }
-          return future;
+          return GenericCompositeFuture.all(futures).mapEmpty();
```
julianladisch, line 268: When running in parallel there can be a race condition in PermsAPI.modifyPermissionArrayField.
If the same permission is granted to two users and the grantedTo property of that permission is updated at the same time both users will read the grantedTo property, the first will correctly update grantedTo, the second will overwrite (revert) the update of the first.

PermsAPI.modifyPermissionArrayField needs to be rewritten to become atomic. A single SQL UPDATE so that we don't have a time slot between a SELECT and then an UPDATE.

```
UPDATE diku_mod_permissions.permissions
SET jsonb = jsonb_set(jsonb, '{grantedTo}', jsonb->'grantedTo' || to_jsonb('33c5ef46-3586-488c-987d-e84116b8eff0'::text))
WHERE id='63351e09-f34c-4aac-a889-c6e8cd7f134e' AND NOT jsonb->'grantedTo' ? '33c5ef46-3586-488c-987d-e84116b8eff0';
```

```
UPDATE diku_mod_permissions.permissions
SET jsonb = jsonb_set(jsonb, '{grantedTo}', (jsonb->'grantedTo') - '33c5ef46-3586-488c-987d-e84116b8eff0'::text)
WHERE id='63351e09-f34c-4aac-a889-c6e8cd7f134e' AND jsonb->'grantedTo' ? '33c5ef46-3586-488c-987d-e84116b8eff0'::text;
```

src/test/java/org/folio/permstest/RestVerticleTest.java
adamdickmeiss: Done, but I can't get the per test timeout to really work ..

src/main/java/org/folio/rest/impl/TenantPermsAPI.java
adamdickmeiss: Ok. This could have been an issue before CompositeFuture.. Not because it's parallel in the CPU sense, but because events are interleaved.. Eg something can happen between "compose" calls. I have no clue on the SQL but it seems to work. Please check my code carefully. Thx.

src/test/java/org/folio/permstest/RestVerticleTest.java
```diff
@@ -44,6 +44,7 @@
 import org.junit.After;
 import org.junit.AfterClass;
 import org.junit.BeforeClass;
+import org.junit.Ignore;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.Timeout;
```
julianladisch, line 50: `io.vertx.ext.unit.junit.Timeout` will allow `Test(timeout = ...)` to overwrite the `@Rule`.

adamdickmeiss: ok

julianladisch (approved):
Looks good, thanks!

PR #166, opened 2022-03-22
Fix loadSample fails on 2nd init MODPERMS-186

adamdickmeiss:
https://issues.folio.org/browse/MODPERMS-186


julianladisch (approved):
Looks good, thanks!

PR #168, opened 2022-03-23
Migrate only users having okapi.all MODPERMS-184

adamdickmeiss:
Set perms.all includes perms.users.assign.{immutable,mutable}.


julianladisch (approved):
Looks good, thanks!

skoczko (approved):

PR #179, opened 2022-05-27
Fix deadlock when creating new user MODPERMS-201

adamdickmeiss:
Do not attempt to obtain separate connection within transaction .. This can cause a deadlock.


julianladisch (approved):
Looks good, thanks!

PR #199, opened 2023-08-07
MODPERMS-220: Delete unused hasPermissions from GET /perms/users API

julianladisch:
The hasPermissions param of the GET /perms/users API has never been implemented.

Removing it is not a breaking change because it can still be passed and still doesn't do anything.

### DIFF
--- a/ramls/permissions.raml
+++ b/ramls/permissions.raml
@@ -49,12 +49,6 @@ traits:
         description: "A query string to filter users based on matching criteria in fields."
         required: false
         type: string
-  byPermission:
-    queryParameters:
-      hasPermissions:
-        description: "A list of permissions that any returned users must possess."
-        required: false
-        type: string
   byMember:
     queryParameters:
       memberOf:
@@ -83,8 +77,7 @@ traits:
         pageable,
         pageable_legacy,
         sortable,
-        queryable,
-        byPermission
+        queryable
       ]
       responses:
         200:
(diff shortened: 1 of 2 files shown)


barbaraloehle:
Hello,
it would be correct to remove the hasPermissions parameter from the 
GET /perms/users API.
The GET /perms/users API contains with sortBy a further parameter which is not implemented.
The GET /perms/permissions API contains the parameters 
memberOf
ownedBy
sortBy
which are not implemented.
All these mentioned parameters were introduced in these APIs by the commit of the file
permissions.raml on Dec 19, 2016 by Kurt Nordstrom,
https://github.com/folio-org/mod-permissions/commit/ac3a59f590a3c81b94427f37618d677d67fe865e .

Instead of removing only the hasPermissions parameter it would be better to remove all unused parameters.
On the other hand there is the question, if there exists the intension to implement these unused parameters timely.

barbaraloehle (approved):
Hello,
it would be correct to remove the hasPermissions parameter from the 
GET /perms/users API.
The GET /perms/users API contains with sortBy a further parameter which is not implemented.
The GET /perms/permissions API contains the parameters 
memberOf
ownedBy
sortBy
which are not implemented.
All these mentioned parameters were introduced in these APIs by the commit of the file
permissions.raml on Dec 19, 2016 by Kurt Nordstrom,
https://github.com/folio-org/mod-permissions/commit/ac3a59f590a3c81b94427f37618d677d67fe865e .

Instead of removing only the hasPermissions parameter it would be better to remove all unused parameters.
On the other hand there is the question, if there exists the intension to implement these unused parameters timely.

PR #206, opened 2024-08-23
MODPERMS-232: DELETE /perms/users/{userId}?indexField=userId

julianladisch:
Add the `indexField=userId` parameter to the existing DELETE API.

https://folio-org.atlassian.net/browse/MODPERMS-232


src/test/java/org/folio/permstest/RestVerticleTest.java
```diff
@@ -401,6 +401,31 @@ public void testDeletePermsUsersByIdDoesNotExist(TestContext context) {
...
+  }
+
+  @Test
+  public void testDeletePermsUsersByUserId(TestContext context) {
+    var id = UUID.randomUUID().toString();
+    var userId = UUID.randomUUID().toString();
+    var permsUsers = new JsonObject().put("id", id).put("userId", userId).encodePrettily();
+    var response = send(HttpMethod.POST, "/perms/users", permsUsers, context);
+    assertThat(response.code, is(201));
+    response = deleteByUserId(id, context);
```
steveellis, line 423: This seems a little confusing. The id is being used in two different ways. One to create the the test data, and then again to cause the 404. Maybe just create another variable called `userIdNotFound` or perhaps just hardcoding a second userid in this line to test the 404.

steveellis: Although I see you have a test dedicated to the 404 test so maybe I'm misunderstanding the purpose of this line.

steveellis (approved):

PR #212, opened 2024-10-29
MODPERMS-233: Rename permissions, add folio-module-descriptor-validator

julianladisch:
https://folio-org.atlassian.net/browse/MODPERMS-233

DO NOT MERGE until all other modules that use a renamed permission or use the bumped interface have a PR ready with the change so that they can be merged at the same time.
* https://github.com/folio-org/mod-users-bl/pull/198
* https://github.com/folio-org/mod-users-keycloak/pull/109

perms.users.item.delete - add subPermissions:
perms.users.item.id.delete
perms.users.item.perm.delete

perms.permissions.get - add subPermissions:
perms.permissions.collection.get
perms.permissions.item.get

Rename perms.permissions to perms.permissions.all. Rename perms.users to perms.users.all.

Add folio-module-descriptor-validator to pom.xml.


descriptors/ModuleDescriptor-template.json
```diff
@@ -54,7 +54,7 @@
             "perms.users.assign.mutable"
           ],
           "permissionsRequired": [
-            "perms.users.item.delete"
+            "perms.users.item.id.delete"
```
steveellis, line 57: Should this be `perms.users.item.perm.delete`? I believe this is the "resource identifier" as specified [here](https://folio-org.atlassian.net/wiki/spaces/FOLIJET/pages/156368925/Permissions+naming+convention) is it not?

steveellis: Ok I see that the `{id}` here is the resource.

steveellis (approved):

PR #216, opened 2024-12-02
MODPERMS-239: Don't deprecate retained split permission ("replaces")

julianladisch:
https://folio-org.atlassian.net/browse/MODPERMS-239

Don't deprecate a permission that still exist.

This allows to split a permission without deprecation so that the old permission is split into old permission and new permission. A user with old permission will have both old and new permission after upgrade.


src/main/java/org/folio/rest/impl/TenantPermsAPI.java
```diff
@@ -341,11 +341,7 @@ private Future<Void> renamePermList(Conn connection, ModuleId moduleId,
             }
           }
           return future;
-        })
-        .compose(cf -> softDeletePermList(permList.values()
-            .stream()
-            .flatMap(List::stream)
-            .collect(Collectors.toList()), connection));
+        });
```
steveellis, line 344: This seems like a small change but I don't understand the implications. Are we wanting to not mark perms as deprecated/replaces anymore at all? [As I said on the ticket](https://folio-org.atlassian.net/browse/MODPERMS-239?focusedCommentId=241354) I don't understand the concept of a "split permission". Perhaps that's the problem.

julianladisch: See the explanation on the ticket https://folio-org.atlassian.net/browse/MODPERMS-239 and the examples in the unit tests of this PR.

julianladisch: No, there can be still permission deprecations.

Before this PR a migration marks permission x as deprecated if permission y has a "replaces": ["x"] entry.

After this PR a migration marks permission x as deprecated if permission y has a "replaces": ["x"] entry __and__ permission x is __not__ listed on its own in the "permissionSets" list.

barbaraloehle (approved):
The solution works, but there is the question of internal implications in the case of the 
splitWithRetain.
The additional class TenantPermsAPITest is very helpful.

src/test/java/org/folio/rest/impl/TenantPermsAPITest.java
```diff
@@ -0,0 +1,220 @@
...
+          "perms": [
+            { "permissionName": "retain.a" },
+            { "permissionName": "retain.x", "replaces": ["retain.a"] },
+            { "permissionName": "retain.y", "replaces": ["retain.a"] }
+          ]
+        }
+        """)
+    .statusCode(201);
+
+    assertPerm("retain.a", "mod-retain");
```
steveellis, line 195: Say that I want to finally remove `retain.a` for the module. Is the process to create a new module version, and not include it in the `perms` list, but also include a `replaces` for it? Does the `permissionRename` test adequately cover this scenario?

julianladisch: To deprecate `retain.a` you remove it from the `perms` list. `replaces` it not needed for deprecation, `replaces` is only used for automatically assign another permission to users.

To delete `retain.a` it must be deprecated and then you need to call `POST /perms/purge-deprecated` API. See RestVerticleTest for unit tests with `/perms/purge-deprecated`: https://github.com/search?q=repo%3Afolio-org%2Fmod-permissions%20purge-deprecated&type=code

julianladisch:
> internal implications in the case of the splitWithRetain

Which use cases are not covered by unit tests?

